PR #30, opened 2020-03-20
Deploy contracts as part of 'make build'

bitwiseguy:
## Description

Fix the `make start` script so that it creates the config files and deploy the contracts before spinning up all containers.

## Related Issue

https://github.com/ethereum-oasis/baseline/issues/29

## Motivation and Context

Fix bug.

## How Has This Been Tested

- deleted my local `<repo_root>/radish34/config/config-*.json` files
- ran `make start`
- ran `make test`
- all tests pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ radish34=./radish34
 # over time, as the radish34 use-case is abstracted/generalized into
 # the baseline protocol, the pushd/popd hacks will fade away...
 
-.PHONY: build clean contracts deploy-contracts npm-install start stop test zk-circuits
+.PHONY: build clean contracts deploy-contracts npm-install start stop rest test zk-circuits
 
 default: build
 
@@ -14,6 +14,7 @@ build: npm-install contracts
 	pushd ${radish34} && \
 	docker-compose build && \
 	npm run setup && \
+	npm run deploy && \
 	popd
 
 clean: stop
@@ -44,6 +45,12 @@ start:
 stop:
 	pushd ${radish34} && \
 	docker-compose down && \
+	popd
+
+reset:
+	pushd ${radish34} && \
+	docker-compose down && \
+	docker volume rm radish34_mongo-buyer radish34_mongo-supplier1 radish34_mongo-supplier2 radish34_mongo-merkle-tree-volume radish34_chaindata && \
 	popd
 
 test:
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ radish34=./radish34
 # over time, as the radish34 use-case is abstracted/generalized into
 # the baseline protocol, the pushd/popd hacks will fade away...
 
-.PHONY: build clean contracts deploy-contracts npm-install start stop rest test zk-circuits
+.PHONY: build clean contracts deploy-contracts npm-install start stop reset test zk-circuits
 
 default: build
 
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,6 @@ build: npm-install contracts
 
 clean: stop
 	docker container prune -f
-	docker network prune -f
 	$(radish34)/../bin/clean_npm.sh
 
 contracts:
@@ -51,6 +50,8 @@ reset:
 	pushd ${radish34} && \
 	docker-compose down && \
 	docker volume rm radish34_mongo-buyer radish34_mongo-supplier1 radish34_mongo-supplier2 radish34_mongo-merkle-tree-volume radish34_chaindata && \
+	npm run deploy && \
+	docker-compose down && \
 	popd
 
 test:
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Here are the targets currently exposed by the `Makefile`:
 | `make npm-install` | `npm ci` wrapper for all modules in the monorepo. |
 | `make start` | Start the full __Baseline__ stack. Requires `docker` service to be running with at least 12 GB RAM allocation. |
 | `make stop` | Stop the running __Baseline__ stack. |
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -6,14 +6,15 @@ radish34=./radish34
 # over time, as the radish34 use-case is abstracted/generalized into
 # the baseline protocol, the pushd/popd hacks will fade away...
 
-.PHONY: build clean contracts deploy-contracts npm-install start stop test zk-circuits
+.PHONY: build clean contracts deploy-contracts npm-install start stop rest test zk-circuits
```
kthomas: Typo :)

bitwiseguy: Fixed

kthomas:
I tried many combinations attempting to make this work.

```
Digest: sha256:66e7d52ef48b18e54e7639c1ab3cd3ff00ecf338ad51c1acd67d613880d4a59f
Status: Downloaded newer image for mongo:latest
Creating radish34_zkp-watch_1 ... done
Creating mongo-buyer          ... done
Creating radish34_zkp_1       ... done

*** Running setup for createMSA ***
curl: (6) Could not resolve host: radish34-zkp.docker
```

Is this the failure you see?

bitwiseguy:
> Is this the failure you see?

I am not seeing that failure. Do you have `dotdocker` running? It is listed in the prerequisites section of `baseline/radish34/README.md`. That service helps resolve the `*.docker` domain names

Makefile
```diff
@@ -46,6 +47,12 @@ stop:
 	docker-compose down && \
 	popd
 
+reset:
+	pushd ${radish34} && \
+	docker-compose down && \
+	docker volume rm radish34_mongo-buyer radish34_mongo-supplier1 radish34_mongo-supplier2 radish34_mongo-merkle-tree-volume radish34_chaindata && \
```
kthomas: I think `make reset` is more aptly named for this.

Can you take a look at `make clean` while you are here and make sure it isn't too aggressive? Meaning... doing things that should be done here in `reset`.

I think `clean` is still a valid target but it should just wipe all the node modules.

bitwiseguy: `make clean` might be a little aggressive in the sense that it will prune ALL docker containers and networks, not just ones related to this `baseline/radish34` project. In order to limit the pruning to this project I think we would have to explicitly list our container names and network names

kthomas:
> > Is this the failure you see?
> 
> I am not seeing that failure. Do you have `dotdocker` running? It is listed in the prerequisites section of `baseline/radish34/README.md`. That service helps resolve the `*.docker` domain names

I installed it but it asked for elevated privileges and I told it to go away... the containers were stopped while pruning containers....

We should the `dotdocker` stuff to the makefile here also.

Makefile
kthomas: For sure. I think we should put all docker-related stuff in `reset` and it should be less aggressive.

And then `make clean` just delegates to npm clean...

bitwiseguy:
> We should the `dotdocker` stuff to the makefile here also.

I wonder if we can get rid of `dotdocker` as a requirement. I don't think it's needed if containers exist within the same docker network. We would just need to make sure inter-container calls use the container names.

If containers  are on the same docker network, you can just use their container names instead of the `VIRTUAL_HOST` env variable name used by `dotdocker`. `dotdocker` is needed if you want to allow communication between containers not on the same docker network, or if you want to communicate from localhost to a container (without having a container port exposed to localhost). 

Thoughts @pmacom @skarred14 @aj-may ?

kthomas:
+1 on dropping `dotdocker` 👍

kthomas:
We also should probably pick a more obscure `--subnet` for the container network. I had to prune my existing networks because they conflicted.

skarred14:
`dotdocker` could be seen as a utility for host based routing, without needing to expose ports on the containers. However, I do see the point of NOT limiting devs to just `dotdocker`. Probably better to use it as an optional setting
@bitwiseguy

kthomas:
> `dotdocker` could be seen as a utility for host based routing, without needing to expose ports on the containers. However, I do see the point of NOT limiting devs to just `dotdocker`. Probably better to use it as an optional setting
> @bitwiseguy

Are there specific performance or security motivations for supporting `dotdocker`? It adds complexity to the setup that seems unnecessary and I don't understand the motivation behind needing to support host-based routing.

kthomas:
I created a branch based on `fix-make-start`: https://github.com/ethereum-oasis/baseline/tree/makefile-cleanup

For posterity...

I am looking into improving the build-and-run lifecycle by reading working copy changes at the module level to determine which container runtimes need to be rebuilt and redeployed, then modifying the `docker` interactions to ensure only changed layers are rebuilt by (i) using `npm i` instead of `npm ci` everywhere and (ii) only performing destructive actions against the docker environment during `make reset`. 

@bitwiseguy

README.md
```diff
@@ -43,6 +43,7 @@ Here are the targets currently exposed by the `Makefile`:
 | `make npm-install` | `npm ci` wrapper for all modules in the monorepo. |
 | `make start` | Start the full __Baseline__ stack. Requires `docker` service to be running with at least 12 GB RAM allocation. |
 | `make stop` | Stop the running __Baseline__ stack. |
+| `make reset` | Clean the docker environment by pruning the docker networks and volumes. |
```
breakpointer: Yeah, effectively "resetting" all user data.

bitwiseguy:
Closing this PR. The quick fix for the readme and makefile are in https://github.com/ethereum-oasis/baseline/pull/42 and there will likely be follow-on refinement from @kthomas

PR #42, opened 2020-03-25
Remove api-watch and zkp-watch containers

bitwiseguy:
## Description

- Remove the `api-watch` and `zkp-watch` containers to simplify and reduce the resources used by our docker environment
- Still using babel + nodemon in the `npm run dev` scripts, but the processes watching for changes to source code are internal to `zkp` and `api-{role}` containers instead of running in separate containers

## How Has This Been Tested

Ran through the entire build and test sequence:
```
make build
make deploy-contracts
make start
make test
```
All tests passed.

- With the containers running, added some `console.log` statements to source code to verify that the hot reload worked in both `zkp` and `api-buyer` containers


Makefile
```diff
@@ -46,6 +44,12 @@ stop:
 	docker-compose down && \
 	popd
 
+reset:
+	pushd ${radish34} && \
+	docker-compose down && \
+	docker volume rm radish34_mongo-buyer radish34_mongo-supplier1 radish34_mongo-supplier2 radish34_mongo-merkle-tree-volume radish34_chaindata && \
+	popd
```
skarred14: probably worthwhile to also prune any stale networks and images as well?

radish34/docker-compose.yml
```diff
@@ -113,13 +100,13 @@ services:
     command: npm run dev
     restart: on-failure
     volumes:
+      - ./api/src:/app/src:delegated
```
skarred14: any reason to map src as well? dist should contain all the compiled js files upon build/restart

radish34/docker-compose.yml
```diff
@@ -148,13 +135,13 @@ services:
...
```
skarred14: same comment as above

radish34/docker-compose.yml
```diff
@@ -183,13 +170,13 @@ services:
...
```
skarred14: same comment as above

radish34/api/package.json
```diff
@@ -17,7 +17,7 @@
   "scripts": {
     "build": "cod-scripts build",
     "build:watch": "npm run build -- --watch",
```
skarred14: is `build:watch` redundant if `dev` includes the same with the change?

radish34/docker-compose.yml
bitwiseguy: I thought this at first too. Since the babel compiler/watcher is running inside the container and watching the `src` directory for changes, it needs to have access to those files. Otherwise, it will not pickup on any file changes in that directory. I tried removing `src` from the mounted directories. It causes an error message `babel: src does not exist` and does not recompile after making changes to the local `src` files.

radish34/docker-compose.yml
bitwiseguy: same response as above

radish34/docker-compose.yml
bitwiseguy: same response as above

radish34/api/package.json
bitwiseguy: Yes, I will remove

Makefile
bitwiseguy: The networks are pruned as part of `docker-compose down`. I added `docker container prune -f` to this command

pmacom:
Hold on merging this PR until it's fully tested. I believe there is a problem with babel now. We are upgrading Node to v13 so we can use import/exports without having to use babel at all.

bitwiseguy:
We will consider upgrading to node v13 in the future

PR #49, opened 2020-03-27
Fix e2e build and cleanup Makefile

kthomas:
# Description

This PR makes the build-and-deploy lifecycle exposed by the `Makefile` incrementally more intuitive (i.e., `make build` produces an IDE-capable workspace, etc.), dramatically improve developer productivity and produce deterministic builds throughout the entire lifecycle.

Also, `dotdocker` is no longer a dependency as a result of these changes.

## How Has This Been Tested

1-- clone the repo and checkout this branch
2-- `make && make start` -- should take a while and work
3-- `make && make start` -- should be very fast this time and work
4-- `make test`

** There are still a few minor failures that we are looking into that appear to be the result of changes made in this branch related to dynamically generating zk material... it does not appear to be getting written in to the buyer or supplier JSON config.

## Types of changes

- [x] Chore
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist


package.json
```diff
@@ -23,7 +23,8 @@
   },
   "scripts": {
     "bootstrap": "lerna bootstrap",
-    "clean": "bin/clean_npm.sh"
+    "clean": "bin/clean_npm.sh",
+    "system-check": "docker system info | awk 'match($0, /Total Memory: (.*)GiB/) {print $3}' | awk '{gsub(\" |GiB\", \"\"); print $0}' | awk '{if ($0 < 12.0) {print \"WARNING: Docker needs at least 12GB to run the baseline stack.\"; exit 1}}'"
```
skarred14: :+1:

radish34/ops/await_stack.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# TODO: wait for all services...
+./ops/await_tcp.sh -t 30 -h localhost -p 8080
```
breakpointer, line 4: I can't help but think we could use dockers built in health check and service dependency graph feature here. What are your thoughts on that? We have a ticket for that approach https://github.com/ethereum-oasis/baseline/issues/51

breakpointer: And potentially some features of swarm.

kthomas: Very much open to that.

This `await_tcp` approach is solid with the exception of needing to add a fallback if `which timeout` fails (will push that fix shortly to make it less annoying for everyone).

Less code to maintain is always better than more. I'd be happy to adopt the docker health check approach you suggest before we merge this.

Will look into #51 in the meantime.

Makefile
```diff
@@ -1,63 +1,90 @@
...
 	docker container prune -f && \
-	docker volume rm radish34_mongo-buyer radish34_mongo-supplier1 radish34_mongo-supplier2 radish34_mongo-merkle-tree-volume radish34_chaindata && \
-	docker network rm radish34_network-buyer radish34_network-supplier1 radish34_network-supplier2 radish34_geth && \
+	docker volume rm \
+		radish34_mongo-buyer \
+		radish34_mongo-supplier1 \
+		radish34_mongo-supplier2 \
+		radish34_mongo-merkle-tree-volume \
+		radish34_chaindata || true && \
+	docker network rm \
```
skarred14, line 68: if there are any active networks on local, this errors out:`ERROR: error while removing network: network ... has active endpoints`. Probably also adding `docker volume prune -f && docker network prune -f && docker system prune -f` as an option here might help

Makefile
```diff
@@ -1,63 +1,90 @@
...
 	docker-compose up -d && \
+	npm run await-stack && \
+	npm run setup-circuits && \
+	npm run deploy && \
 	popd
 
 stop:
-	pushd ${radish34} && \
+	@pushd ${radish34} && \
 	docker-compose down && \
```
skarred14, line 53: `docker-compose down --remove-orphans`

Makefile
```diff
@@ -1,63 +1,90 @@
...
-	pushd ${radish34} && \
-	docker-compose build && \
-	npm run setup && \
+
+build-containers: build
+	@pushd ${radish34} && \
+	npm run dockerize && \
 	popd
 
 clean: stop
```
skarred14, line 20: 👍

radish34/ops/build_api.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
skarred14, line 1: this is definitely a great improvement! `pushd` doesnt work the way it was invoked in `package.json` on ubuntu, unless it is explicitly specified in `.npmrc` - this solves it better

radish34/ops/await_tcp.sh
```diff
@@ -0,0 +1,181 @@
+#!/bin/bash
+
+WAITFORIT_cmdname=${0##*/}
```
skarred14, line 3: 💯

Makefile
kthomas: Fixed

kthomas:
All suites green. Thanks for the assist @bitwiseguy 🎸

We can merge this 🚀

@skarred14 @breakpointer @pmacom

skarred14:
All tests pass

kthomas:
image

UI loading as of `017a1408b124212caef63d7b9e3769aa506c248f`.

PR #56, opened 2020-03-28
Enhancements/code/cleanup

balajipachai:
# Description

- General code cleanup.
- Removed await from places where it was not needed.
- Introduced parameter destructuring.
- Added comments on top of the functions 

## Related Issue

- N / A

## Motivation and Context

- Following code destructuring enhances code readability, in turn, makes life easy for new developers to walk through the code.

### DIFF
--- a/radish34/messenger/src/clients/whisper/whisperUtils.js
+++ b/radish34/messenger/src/clients/whisper/whisperUtils.js
@@ -25,13 +25,14 @@ async function sendPrivateMessage(
     messageString = JSON.stringify(messageContent);
   }
   const web3 = await web3utils.getWeb3();
-  const content = await web3.utils.fromAscii(messageString);
+  const content = web3.utils.fromAscii(messageString);
   const whisperId = await Identity.findOne({ _id: senderId });
+  const { keyId } = whisperId;
   let hash;
   try {
     hash = await web3.shh.post({
       pubKey: recipientId,
-      sig: whisperId.keyId,
+      sig: keyId,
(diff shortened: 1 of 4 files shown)


skarred14:
lgtm

radish34/messenger/src/clients/whisper/whisperUtils.js
```diff
@@ -44,7 +45,7 @@ async function sendPrivateMessage(
   }
 
   // Store message in database
-  const time = await Math.floor(Date.now() / 1000);
+  const time = Math.floor(Date.now() / 1000);
```
pmacom: 👍

radish34/messenger/src/utils/generalUtils.js
```diff
@@ -14,6 +14,10 @@ const {
 
 const radishApiUrl = process.env.RADISH_API_URL ? `${process.env.RADISH_API_URL}/api/v1` : 'http://localhost:8101/api/v1';
 
+/**
+ * Function that checks whether given object has a JSON structure?
+ * @param {Object} str
```
bitwiseguy: `str` should be type `String`. This function attempts to convert that string back into an object. There is some extra complexity to catch some edge cases for `str === "true"` and `str === "null"`. Can we change the comment to the following?

```
/**
 * Checks whether a given string can be converted to a proper JSON object.
 * Edge cases: inputs of "true" or "null" return false
 * @param {String} str 
 */
```

PR #90, opened 2020-04-02
Removed old manager param as it was misleading and not needed

Perseverance:
# Description

Changed the Registrar contract to pass itself as address of the contract that is managed by the manager instead of passing it from the outside. Modified the tests

## Motivation and Context

The parameter was not needed and the param name and description were misleading. Minor Gas optimization also

## How Has This Been Tested

Unit tests

## Types of changes

## Checklist

### DIFF
--- a/radish34/.gitignore
+++ b/radish34/.gitignore
@@ -101,3 +101,4 @@ config/merkle-tree/
 
 # Ignore generated docs
 docs/generated
+.vscode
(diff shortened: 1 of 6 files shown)


radish34/.gitignore
```diff
@@ -101,3 +101,4 @@ config/merkle-tree/
 
 # Ignore generated docs
 docs/generated
+.vscode
```
skarred14, line 104: minor update: line break needed here

Issue #95, opened 2020-04-04
npm run deploy - Error: incorrect number of arguments


jeffprestes:
## Expected Behavior
Assigned the deployer as the manager for OrgRegistry. TxHash: 0x1689ac60fba5c25b5559e0fbca066c4063a433dec552adc9208977e161f05852

## Actual Behavior
Patiently waiting 10 seconds for ganache container to init ...
Checking for ganache ...
(node:9) UnhandledPromiseRejectionWarning: Error: incorrect number of arguments
    at Contract.assignManager (/app/node_modules/ethers/contract.js:130:19)
    at Object.assignManager (/app/src/utils/organization.js:29:39)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async assignManager (/app/src/deploy.js:39:31)
    at async main (/app/src/deploy.js:115:3)

## Steps to Reproduce the Problem

  1. Follow setup instructions until step 6

## Specifications

  - Version: latest
  - Platform: MacOs 10.15.4
  - Subsystem: node version v11.15.0

bitwiseguy:
@jeffprestes I have not seen this error before. Are you using the latest code from the `master` branch and have you customized it in any way? Here are a few things to consider that could help solve the issue:

1. Clean out the blockchain data by removing that docker volume:
```docker volume rm radish34_chaindata```
2. Rebuild the `deploy` container to ensure it is not broken:
```docker-compose build deploy```
2. If you see the same error again, can you share the logs from the `ganache` container?
```docker logs ganache```

**Comment by jeffprestes** [comment]
Error persists. Looks like ganache container is not loading:

```bash
$ docker logs ganache
Error: No such container: ganache
``` 
Is there a way to start ganache container manually? 
Also, use GETH instead of Ganache would be safe, wouldn't be?

skarred14:
@jeffprestes good catch! thanks. Recent merge from #90 seemed to have caused this issue. Take a look at #97 which should fix this bug

PR #97, opened 2020-04-06
fix(bug): deployment script changes for manager assignment

skarred14:
# Description

Changes in this PR fix issue with `npm run deploy`, where it fails during the call: 
` const { transactionHash } = await Organization.assignManager('OrgRegistry', role);`. This error is due to the recent change in the number of parameters in the `OrgRegistry.assignManager(..)` in #90 
## Related Issue

#95 

## Motivation and Context

Fix bug to get the demo app on master to work.

## How Has This Been Tested

All tests including `npm run test` pass

## Types of changes

## Checklist


Makefile
```diff
@@ -49,6 +49,7 @@ reset:
 	docker-compose down && \
 	docker container prune -f && \
 	docker volume rm radish34_mongo-buyer radish34_mongo-supplier1 radish34_mongo-supplier2 radish34_mongo-merkle-tree-volume radish34_chaindata && \
+	docker network rm radish34_network-buyer radish34_network-supplier1 radish34_network-supplier2 radish34_geth && \
```
kthomas, line 52: 👍

radish34/bin/setup.sh
```diff
@@ -9,7 +9,7 @@ printf "\n${GREEN}*** Starting zokrates container ***${NC}\n"
 docker-compose up -d zkp
 
 # delay needed to ensure all container are in running state.
-sleep 5
+sleep 10
```
breakpointer, line 12: 🤦‍♂️I could use more sleep... computers should not.

radish34/contracts/__tests__/mockVerifierData.js
```diff
@@ -9,62 +9,135 @@
...
-        "0x020ca55394bdb87edfbcb54e6961cac54f83167d7e1574e60b00d7e0ffedeac4" ],
-    "proofs_int": ["14989707767102615313684605173057903641347878660550199499377470094238716936181",
-        "10575566526490722493368634185430526475607354877647401472809285966069569617931",
-        "3784626748982707459523154788722591507947491336303157290452437655043216152955",
-        "8183624607866254943801646970430624514997174711794003680811213610669243946554",
-        "709630340745687344505736606096953814072381190831471097905206887612184446377",
-        "689451577733667459872895366744634263233388447995637921789623037416555818880",
-        "21130479047924098851097522361915995958233452872465264954591872849418039442457",
-        "926968903430932935510549259338269173796291853932043469374211440611536267972"],
+  "msa": {
```
breakpointer, line 12: Why did these change?

radish34/contracts/src/utils.js
```diff
@@ -77,3 +77,114 @@ export const link = (bytecode, libraryName, libraryAddress) => {
   // swap out the placeholder with the library's deployed address:
   return bytecode.object.replace(pattern, address);
 };
+
+export const flattenDeep = arr => {
```
breakpointer, line 81: Were these somewhere else before?

radish34/bin/setup.sh
kthomas: > 🤦‍♂️I could use more sleep... computers should not.

lol. +1

radish34/api/src/services/po.js
```diff
@@ -335,6 +335,9 @@ export const createPO = async (_zkpPrivateKeyOfBuyer, oldMSA, newMSA, po) => {
   } = await createPOTransaction(proofArray, publicInputHash.field(248, 1, true), publicInputs);
 
   console.log('\nCREATE PO TRANSACTION COMPLETE');
+  console.log('proofArray:', proofArray);
```
skarred14, line 338: these console.logs need to be removed

radish34/contracts/__tests__/mockVerifierData.js
skarred14: during testing, we needed to generate new sets. ideally, we could something like `axios` to mock this data

radish34/contracts/src/utils.js
skarred14: yes these were in `/api/src/utils/crypto/conversions.js` these were meant to be used to convert hex to int. We can get rid of these from here. ideally, we should be making any such utils as part of the package

radish34/deploy/src/deploy.js
```diff
@@ -109,7 +116,7 @@ const saveSettings = async role => {
 };
 
 const main = async () => {
-  const registerAll = process.env.MODE === 'register-all';
+  // const registerAll = process.env.MODE === 'register-all';
```
skarred14, line 119: lets remove this line

radish34/contracts/src/utils.js
kthomas: I similarly had to duplicate some of the babyjubjub code within the `deploy/` package-- for now. Let's punt DRYing up utils to the package in the spirit of getting to stability and CI.

radish34/bin/setup.sh
bitwiseguy: Every 5 seconds of sleep is valuable! I think this should be replaced by polling a healthcheck on the container or something similar. There is an open issue for adding container healthchecks: https://github.com/ethereum-oasis/baseline/issues/51

radish34/contracts/__tests__/mockVerifierData.js
bitwiseguy: Previously this file only contained mock data for the `createMSA` transactions. I added some nesting so that the data is separated for `msa` and `po`. So the changes to this file are to nest all the existing data in an `msa` object and add similar data in a new `po` object

radish34/api/src/services/po.js
bitwiseguy: Removed

radish34/deploy/src/deploy.js
bitwiseguy: Removed

radish34/contracts/src/utils.js
bitwiseguy: I left them in but added a `TODO` comment explaining that they are duplicates and we intend to make them part of an importable package in the future.

radish34/api/src/services/po.js
breakpointer: Well, yes probably. If they're helpful I could see leaving them in for now. However it does seem like this is sensitive information (or at least would be in a production setting). We do need to do a security review of our logs at some point (also convert them to proper logging instead of `console.log` where we can make these specific logs `DEBUG` only logs. Here's a ticket I wrote for that work effort. https://github.com/ethereum-oasis/baseline/issues/101

radish34/contracts/__tests__/mockVerifierData.js
breakpointer: okay thanks for the explanation 👍

radish34/bin/setup.sh
breakpointer: Yeah, I'm looking at the healthcheck issue today.

radish34/api/src/services/po.js
bitwiseguy: Yes good point. These logs might be a good candidate to display in "debug" mode but not "production" mode

Issue #88, opened 2020-04-02
Story: Tidying up the contracts deployment procedure and preparing it for baseline


Perseverance:
The current contracts deployment procedure for Radish34 is based on a nodejs project and custom deployments of several contracts. Although it works flowlessly for an MVP procedure, several improvements can be made to make it "baseline" ready:

### Separating and abstracting away, the deployment processes for client contracts and protocol contrats.

Current contracts deployment procedure is unified for every necessary contract in order to run Radish34 locally. Looking forward into abstracting this procedure, so that it becomes a part of baseline, the contracts can be separated in two categories.

- Protocol contracts - contracts that will be present on mainnet at the time of deployment and do not need to be deployed, but need to be deployed when using local environment. Ex. ERC1820Registry contract.
- Client contracts - contracts that will need to be deployed (and maintained - in future story) as part of the client setup - separate instance for every client. Ex. Org Registry, Shield, Verifier

We can create a mini-library that can be used to execute the procedure for both categories of contracts. This library will allow for configuring contracts, picking and choosing what to be deployed and generally perform the deployment process given the inputs. 

This library will become the core of the next story.

### Tidying up the code for better readability and maintenance

Currently the code uses custom wrappers around ethers and native ethers.js code for deployment of contracts. Most of these are already present in other ethers.js based deployment libraries like [etherlime-lib](https://etherlime.gitbook.io/etherlime/developer-documentation/etherlime-library-api/deployers). These will allow for reducing and tidying the codebase.

Perseverance:
@breakpointer @skarred14 Can you guys assign me to this issue?

skarred14:
Thanks for submitting this issue. Quick comment about ERC1820 in Protocol Contracts:
ERC1820Registry is already deployed on the mainnet at: `0x1820a4B7618BdE71Dce8cdc73aAB6C95905faD24`. For a mainnet deployment, minor changes are necessary in the current sol file constructors, to update this value, instead of re-deploying ERC1820Registry. Considering other alternative networks (Ropsten. Ganache, etc.) as targets could be a good aspect of the mini lib

Perseverance:
@skarred14 Noted, I will account for that.

Perseverance:
Closed by #109

PR #98, opened 2020-04-06
Access restriction to the OrgRegistry registerOrg and registerInterfaces methods

Perseverance:
# Description

Changed the Radish34 OrgRegistry, Verifier and Shield contracts to have the ability to have multiple owners. Access restricted registerOrg and registerInterfaces as currently everyone can call them and "join" the workgroup.

This is needed in order to create a more streamlined deployment/administration procedure.

## Related Issue
Loosely Related to #88 

## Motivation and Context

The current implementation of OrgRegistry allows anyone to register a new organization or interface. This should only be possible by the workgroup admin. As there probably can be cases of more than one organisations being administrators of the workgroup, a multi-ownership scheme is added.

## How Has This Been Tested

This has been tested via the existing and new unit tests. The deployment procedure for Radish34 has been tested via `npm run deploy` expecting the same results as previously.

## Types of changes

## Checklist


radish34/contracts/__tests__/OrgRegistry.test.js
```diff
@@ -59,12 +59,45 @@ test('Should be able to register an org', async () => {
...
+  try {
+    await nonOwnerOrgRegistry.registerOrg(
+      accounts[1],
+      utils.formatBytes32String("Name"),
+      2,
+      utils.hexlify('0x0471099dd873dacf9570f147b9e07ebd671e05bfa63912ee623a800ede8a294f7f60a13fadf1b53d681294cc9b9ff0a4abdf47338ff72d3c34c95cdc9328bd0128'),
+      utils.hexlify('0x0471099dd873dacf9570f147b9e07ebd671e05bfa63912ee623a800ede8a294f7f60a13fadf1b53d681294cc9b9ff0a4abdf47338ff72d3c34c95cdc9328bd0128')
+    );
+  } catch(e) {
+    expect(e.message).toMatch('revert');
```
skarred14, line 78: It would be better to throw an exception from smart contract, to then catch an exception here to check `e.code` and `e.message` matches the error message in the require condition within the contract

radish34/contracts/__tests__/OrgRegistry.test.js
```diff
@@ -59,12 +59,45 @@ test('Should be able to register an org', async () => {
     utils.hexlify('0x0471099dd873dacf9570f147b9e07ebd671e05bfa63912ee623a800ede8a294f7f60a13fadf1b53d681294cc9b9ff0a4abdf47338ff72d3c34c95cdc9328bd0128'),
     utils.hexlify('0x0471099dd873dacf9570f147b9e07ebd671e05bfa63912ee623a800ede8a294f7f60a13fadf1b53d681294cc9b9ff0a4abdf47338ff72d3c34c95cdc9328bd0128')
   );
+  const orgCountBefore = await orgRegistry.getOrgCount();
+  expect(orgCountBefore.toNumber()).toBe(1);
+});
+
+test('Should not be able to register an org from non-owner', async () => {
+  const nonOwner = await getSigner(accounts[4]);
```
skarred14, line 67: you could create a random wallet using ethers

radish34/contracts/__tests__/OrgRegistry.test.js
```diff
@@ -59,12 +59,45 @@ test('Should be able to register an org', async () => {
...
   const txReceipt = await orgRegistry.registerInterfaces(utils.formatBytes32String("RandomInterface"), accounts[0], accounts[1], accounts[2]);
 });
 
+test('Should not be able to register an orgs interfaces from non owner', async () => {
+  const nonOwner = await getSigner(accounts[4]);
+  const nonOwnerOrgRegistry = orgRegistry.connect(nonOwner);
+  try {
+    await nonOwnerOrgRegistry.registerInterfaces(utils.formatBytes32String("RandomInterface2"), accounts[0], accounts[1], accounts[2]);
+  } catch(e) {
+    expect(e.message).toMatch('revert');
```
skarred14, line 95: same comments as above

radish34/contracts/contracts/MultiOwnable.sol
```diff
@@ -0,0 +1,30 @@
...
+     * @dev Function to set owners addresses
+	 * @notice change this to DAO like governence as this is too exploitable. Probably good enough for Radish34 for now
+     */
+    function setOwner(address owner) public {
+        require(isOwner[msg.sender]);
+        isOwner[owner] = true;
+        emit SetOwner(owner);
+    }
+
+}
```
skarred14, line 30: add line break here

radish34/deploy/src/utils/organization.js
```diff
@@ -22,11 +22,9 @@ const assignManager = async (manageeName, newManagerName, msgSenderName) => {
     Wallet.getPrivateKey(msgSenderName),
```
skarred14, line 22: as discussed, this would be modified when rebased against #97

radish34/contracts/contracts/MultiOwnable.sol
skarred14: also, is there a separate test case for `Multi-Ownable`?

radish34/contracts/__tests__/OrgRegistry.test.js
Perseverance: My reasoning behind using one of the unused funded wallets, in order to skip the need of sending some Ganache ETH to the newly generated wallet. Do you think it is necessary to generate a brand new one, or maybe I can just bump it up to the last address account[9]?

radish34/contracts/contracts/MultiOwnable.sol
Perseverance: Good catch. I tested it through the OrgRegistry but I will add a separate one for MultiOwnable

radish34/contracts/__tests__/OrgRegistry.test.js
skarred14: makes sense, its fine to keep it this way to use an already existing account in the wallet for tests like these

radish34/contracts/__tests__/OrgRegistry.test.js
Perseverance: Addressed

radish34/contracts/__tests__/OrgRegistry.test.js
Perseverance: Addressed

radish34/contracts/contracts/MultiOwnable.sol
Perseverance: Addressed

radish34/contracts/contracts/MultiOwnable.sol
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * @dev Function to set owners addresses
+	 * @notice change this to DAO like governence as this is too exploitable. Probably good enough for Radish34 for now
+     */
+    function setOwner(address owner) public {
+        require(isOwner[msg.sender], "Unauthorised access: needs to be called by an owner!");
+        isOwner[owner] = true;
+        emit SetOwner(owner);
+    }
+
```
breakpointer, line 29: How do we remove owners? Can one owner remove all other owners? I have a lot of related questions about the 'ownership' lifecycle. Are there existing standard contracts that we could use here instead? I like the idea of supporting this but I think there are a lot of corner cases to think through.

Perseverance:
> I appreciate the effort to make this update and conceptually I think this an interesting change. However, I have some concerns are around the implications of this change to the secure setup process and individual participant onboarding. Specifically, does this change now require parity of 'owners' of this registry contract and those that participated in the trusted setup of zkp? Also, after the setup phase is complete, at some point in the future a new organization wants to become an 'owner' of this contract as well. What is that process like?
> 
> Somewhat tangental, but does adding multiple owners here increase the security threat surface of the contract? (e.g. if one owner of the many owners wallet is security is compromised...)
> 
> It would be great to have @skarred14 and @LucasRodriguez review and help think this change through.

Thanks for the thorough comment @breakpointer ! Let me clear the motivation behind this PR. My main goal with this PR has been to access restrict the `registerOrg` and `registerOrgInterface` methods. These, are currently free to call by anyone which given enough sophistication can be abused.  I realize that this solution is far from ideal for prod, or even test version.

According to my understanding, this change does not change the secure setup process. As for the increased attack vector for the multiple owners - I totally agree. 
My thinking for changing the single to multiownership model has been that the Radish34 demo case is used for demonstration purposes and while demoing this, the demonstrator might want to showcase that there is no single organization in the workgroup that has more privilege (being the single owner) than the others. I am totally fine reverting it to single ownership if you guys think that this is too far fetched.

In summary, my main goal has not been to make a breakthrough in the governance model of a workgroup - I think this is something for us to discuss and architecture in much greater depth later. I've made a makeshift solution where one/multiple companies can be made owners so that we access restrict how we add more organizations. I fully realize that this will be likely changed a lot when we figure out the organization's governance model.

radish34/contracts/contracts/MultiOwnable.sol
Perseverance: I did a lot of contemplation of whether we should have functionality removing of owners. My (maybe dumb) reason was that if we do not have removal of organizations in this version, probably we won't need to remove owners either. Again that might be a super dumb argument 😂

LucasRodriguez: I cant see anything right now that:
1. Lists all of the owners
2. Returns a count of the number of owners. 

I think we should consider adding this, just in case.

Adding to @breakpointer 's question,
We can't/shouldn't have a multi-owner with out having an optional _super-owner_ to resolve conflicts. Otherwise we might eventually find ourselves with governance issues over the contract. 

**Questions to address around this IMO:**
- Who can delete another user? 
- Should you be able to delete another user at all?
- Should any owner have the ability to add any other owner?
- Can the same functionality be added with something that doesn't mess with ownership? a whitelist-mechanism perhaps?

Perseverance: Yep, most of these questions are governance related, which probably is a whole different subject by itself. We can go ahead with the multi  owner or even revert to single owner. I guess governance discussion call is needed @skarred14

LucasRodriguez:
AFAIK, the `verifier` and `shield` don't require multi-ownership. I see where this is coming from, but imo it doesn't solve any issues. It does solve a potential optics problems, and tries to address it by sharing the ownership amongst participants. However these are public smart contracts, and all of the information is available to all participants. Functionally having many owners doesn't add any functionality to the project. Beyond that, this introduces governance issues and an increase in attack vectors, as previously mentioned.

Now, Regarding the `registerOrg` and `registerOrgInterface`. I think that the improvements proposed can be addressed in ways that don't mess with the governance of the contracts, adding several owners raises more questions that it solves problems. I'd be happy to either discuss other methods or take the time to collaboratively answer all the necessary questions.

Perseverance:
> AFAIK, the `verifier` and `shield` don't require multi-ownership. I see where this is coming from, but imo it doesn't solve any issues. It does solve potential optics problems and tries to address it by sharing ownership amongst participants. However these are public smart contracts, and all of the information is available to all participants. Functionally having many owners doesn't add any functionality to the project. Beyond that, this introduces governance issues and an increase in attack vectors, as previously mentioned.
> Now, Regarding the `registerOrg` and `registerOrgInterface`. I think that the improvements proposed can be addressed in ways that don't mess with the governance of the contracts, adding several owners raises more questions than it solves problems. I'd be happy to either discuss other methods or take the time to collaboratively answer all the necessary questions.

I'd love to discuss other methods, in the mean time do you consider single ownership a less damaging solution?

skarred14:
agree to @LucasRodriguez point, that it may yet be a bit early to have shield and verifier contract functionality restricted. They are intended to be public anyway. However, if governance requirement for a given use case deems that certain functions are to be restricted, then that could be managed through OrgRegistry. It would be good to validate this in the larger context of access management as a separate feature request. Let's have a discussion on this to decide on the core functionalities when it comes to governance and additional privacy in smart contract management.

@Perseverance there is value in having this deployable cli-util re-usable any other entity in general. Do you have any suggestions on how the `MultiOwner` could be modified without touching upon governance as part of this PR?

Perseverance:
> agree to @LucasRodriguez point, that it may yet be a bit early to have shield and verifier contract functionality restricted. They are intended to be public anyway. However, if governance requirement for a given use case deems that certain functions are to be restricted, then that could be managed through OrgRegistry. It would be good to validate this in the larger context of access management as a separate feature request. Let's have a discussion on this to decide on the core functionalities when it comes to governance and additional privacy in smart contract management.
> 
> @Perseverance there is value in having this deployable cli-util re-usable any other entity in general. Do you have any suggestions on how the `MultiOwner` could be modified without touching upon governance as part of this PR?



> agree to @LucasRodriguez point, that it may yet be a bit early to have shield and verifier contract functionality restricted. They are intended to be public anyway. However, if governance requirement for a given use case deems that certain functions are to be restricted, then that could be managed through OrgRegistry. It would be good to validate this in the larger context of access management as a separate feature request. Let's have a discussion on this to decide on the core functionalities when it comes to governance and additional privacy in smart contract management.
> 
> @Perseverance there is value in having this deployable cli-util re-usable any other entity in general. Do you have any suggestions on how the `MultiOwner` could be modified without touching upon governance as part of this PR?

I've thought about it and I think all of you are right. MultiOwnable is essentially a governance change. If it is ok with you, I will revert the contracts to single ownership and discuss the Governance in another issue. 
Only thing in this PR would be to make the registering organization and interfaces to be `onlyOwner` (bound to the single owner).

Does this sound Ok to you @LucasRodriguez @skarred14 @breakpointer ?

skarred14:
In general, the idea of making deployment a cli-util is a good choice, so that the deployment can be run by any user. Would adding onlyOwner() as above help in doing so?

kthomas:
Would setting the owner to a multisig address effectively delegate governance to a group of owners and maintain clear separation of concerns? In other words, could that pattern be a way for us to punt on adding any governance functionality within the OrgRegistry itself for now?

Perseverance:
> In general, the idea of making deployment a cli-util is a good choice, so that the deployment can be run by any user. Would adding onlyOwner() as above help in doing so?

It wont make too much of a difference I guess. It just needs to be access restricted so that it is not abused.

> Would setting the owner to a multisig address effectively delegate governance to a group of owners and maintain clear separation of concerns? In other words, could that pattern be a way for us to punt on adding any governance functionality within the OrgRegistry itself for now?

I think this idea is wonderful! I spoke briefly with the Zeppelin Guys sometime ago and they seem to support this style too

Perseverance:
PR Updated - Reverted to single owner in order not to change governance.

LucasRodriguez:
Agreed on reverting to sole ownership.

@Perseverance could you clarify for where we stand regarding the restrictions to `Verifier` and `Shield`. I would like to reiterate any ownership modifiers to the functions in these contracts can potentially conflict with their over-arching objective, or not add any value at the cost o more loc. 

> Would setting the owner to a multisig address effectively delegate governance to a group of owners and maintain clear separation of concerns? In other words, could that pattern be a way for us to punt on adding any governance functionality within the OrgRegistry itself for now?

I would delegate this conversation for now. See if we can open a discussion around governance and  subsequent patterns to deal with it effectively, before we commit to any solutions

Perseverance:
> Agreed on reverting to sole ownership.
> 
> @Perseverance could you clarify for where we stand regarding the restrictions to `Verifier` and `Shield`. I would like to reiterate any ownership modifiers to the functions in these contracts can potentially conflict with their over-arching objective, or not add any value at the cost o more loc.
> 
> > Would setting the owner to a multisig address effectively delegate governance to a group of owners and maintain clear separation of concerns? In other words, could that pattern be a way for us to punt on adding any governance functionality within the OrgRegistry itself for now?
> 
> I would delegate this conversation for now. See if we can open a discussion around governance and subsequent patterns to deal with it effectively, before we commit to any solutions

The access restrictions for Shield and Verifier were in place already before this PR. It is only that I access restricted the assignManager method so that no arbitrary person can trigger it. Do you think this one is causing trouble?

LucasRodriguez:
AFAIK, anyone should be able to call the `verifier` functions. In terms of the `shield`, it  has its own safeguards to ensure that no one can modify its contents or add anything unless they are following a set of rules set out beforehand.

IMO any arbitrary person should be able to call both, for the `verifier` to be restricted would miss the point and the `shield` already ensures that even if a random person calls it, nothing gets added so I believe this to be redundant code

thoughts @skarred14 ?

Perseverance:
> AFAIK, anyone should be able to call the `verifier` functions. In terms of the `shield`, it has its own safeguards to ensure that no one can modify its contents or add anything unless they are following a set of rules set out beforehand.
> 
> IMO any arbitrary person should be able to call both, for the `verifier` to be restricted would miss the point and the `shield` already ensures that even if a random person calls it, nothing gets added so I believe this to be redundant code
> 
> thoughts @skarred14 ?

Unfortunately I am not sure I completely understand the issue here. Both Shield and Verifier are completely free to call in terms of verifying. Only addition, as seen in the code, is access restricting the `assignManager` method which is not part of the main business logic but more or less maintenance for the erc1820.

PR #109, opened 2020-04-12
Radish deployment library

Perseverance:
# Description
Refactors the deployment process in a library. The library will serve as basis for the creation of several tools on the roadmap. This PR had the goal of refactoring the code while outputting the same results as previously in order to keep the demo from any breaking changes.

## Architectural description

The Radish contracts deployment and management library will be used as basis for the Baseline deployment and management library and the baseline-administrator tools. Here is an overview of the library and several Radish specifics.

Radish Deployment & Management Library Architecture

### Setting the scene
The idea behind the library is to act as a foundation for generation of tools used for baseline deployment and management. The current version is fitted for the needs of Radish34 but has been built with extensibility in mind (without overengineering it based on the project youth). The main user of this library is the workgroup administrator allowing her/him to deploy the needed contracts and manage the workgroup through them.

### Logical Input Objects

There are 4 (+1 included in the last section for Radish specific config) main logical input objects. These are needed in order for a workgroup to be created. These are:
- Contract Artifacts - artifacts needed for the deployment of the contracts. At a minimum, these should have the contracts bytecode, name and abi. In Radish34 these are taken from the artifacts produced from the compilation.
- Keystore or Key Provider - object allowing a wallet to be derived out of it. This wallet will be used for further blockchain network interaction. In Radish34 these are taken from the keystore directory
- Organisation Config - object carrying organization-specific information. At a minimum, these should have the organization name, role and zero-knowledge private key. In Radish34 these are taken from predefined configuration files.
- Verification Key - object carrying a verification key and type for a specific ZK circuit. In Radish, these should be requested from the zkp service.
### Resolvers
All main logical input objects can have multiple ways to be inputted. In order to allow for new ways of resolving these objects, the notion of Resolvers is introduced. As long as Resolvers follow a similar interface, the "action" classes will be able to use them to resolve the necessary logical input objects. In the Radish34 demo, each logical object has it's specific resolver based on the source of the object.

For Contract Artifacts, it is based on a .json configuration file. For Keystore, it is based on the directory and the name of the organization. For Organisation it is based on directory and the name of the organization leading to a config file. For the verification key, it is based on a http request.

Using this approach, new resolvers can be defined and included in the library, which will allow for better extensibility. This would be beneficial, especially for resolving sensitive data (like private keys) remotely;

### Baseline Deployer
The baseline deployer is an action class allowing its user to deploy the current baseline contracts (currently these are only the Radish34 contracts).

The deployer expects an ethers.Signer object and ethers.Provider object as creation parameters. These will be used in all subsequent deployments.
At the moment, two types of contracts can be logically outlined - protocol contracts and workgroup contracts.

The protocol contracts are contracts that are expected to be already in place in the public networks. For local development, these need to be deployed. An example of this is the ERC1820Registry contract that is available on Mainnet, but not on your local ganache.

The workgroup contracts are contracts specific for the workgroup being created. These will be deployed on any network you are using - mainnet or local. Examples of these contracts are the OrgRegistry, Shield and Verifier.

#### Builders
The deployment procedure is aided by two builder classes - for the two contract types. These builder classes allow the administrator to specify which parts of the deployment process to be executed. This will come especially handy when extending this library to different Baseline tools.

Once the builders are used to create deployment tasks, these tasks are submitted to the deployer object and the deployer returns the deployed contract instances.

### Baseline Workgroup Manager
The baseline workgroup manager class is an action class allowing its users to manage a workgroup. At present times, a workgroup is defined by three contracts - OrgRegistry, Shield, and Verifier. Once these are supplied (connected to the intended transaction sender), the workgroup manager can start managing and configuring the workgroup.

Current abilities of the workgroup manager are:
- Registering organizations
- Registering organization interfaces
- Registering verification keys

### Radish specific config 
One additional logical object specific to Radish is defined. This object defines the different organizations that the Demo will be bootstrapped with. 

Outside of the demo context of Radish34, the processes of deployment and administration will probably be detached. This means that this configuration object will probably be passed separately for each organisation. The current version is only tailored for Radish34.

## Related Issue

Resolves to #88 

## How Has This Been Tested

The whole deployment flow and running has been tested on Ganache and via the Radish34 deployment procedure.

Deployment Result

## Types of changes

## Checklist


LucasRodriguez:
If I understand the purpose of the PR correctly, the objective is to abstract the baseline contract deployment into a library. This library should also provide the ability to extend the baseline to accommodate for more complex projects.

The PR should, therefore, focus on how it will achieve this abstraction and extensibility, leaving behind how the process works in radish.

We should build this streamline of baseline independently of the radish use-case. I understand the difficulty of doing so, but building something generic from the base of a use-case will create technical debt, and we will have to come back and refactor this eventually.

Diving further into the PR I have several comments:

### Logical Input Objects
I think we should revisit the structure of this section like so:
**Input:** Name of the input 
**Justification:** What purpose it serves, and why it is a baseline input
**Input type:** 
 - *System Store* - It is required for the initialization of the system but is, explicitly, not an extensible input
 - *User Input* - The user determines the input and passes the parameters to the library on init
 - *Hybrid* - A baselined input comes by default, and is required for the initialization, but can be extended by the user
**Structure:** A description of how the input will be fed to the library, and in what format
**Reference:** A mention of what is currently being used in radish34 to help us map the functionality and process. 

An example would look like: 
> **Input:** Contract Artifacts
> **Justification:** Any baseline project will require the initialization of some basic contracts defined in this input (potentially even list the basic contracts? )
> **Input type:** *Hybrid* - A baselined input comes by default, and is required for the initialization, but can be extended by the user
> **Structure:**  At a minimum, these should have the contracts bytecode, name and abi.
> **Reference:** In Radish34 these are taken from the artifacts produced from the compilation.

@Perseverace I know your description contains most of these already but adding the structure and extra elements will help us form a better picture of how we will be abstracting the code. This will turn out to be a huge effort, and clarity will  aid in the review. 


### Resolvers
I am struggling to get a good grasp on what resolvers are. So please correct if I'm wrong The resolvers work as a middleware to translate the input given by the user to the input accepted by the library. While I think to see the usefulness of having this functionality, I think we are much better off hardcoding the type of the inputs instead of giving the user that flexibility. This can become very difficult to maintain with time, perhaps I'm missing something but this seems to be something the user could do without that doesn't contribute to the concept of “baselining”


### Baseline Deployer
This seems good overall, I would include an idea of how key management is going to be handled or explicitly leave that out of the scope of this library.
We should also consider whether we want a baseline library for each part of the baseline or a single library that builds a baseline application from scratch. There are other parts too that require deployment and abstraction in a similar way we should discuss how we plan to bring, or not bring, them together.
This suggests that we might be looking at two different kinds of deployments, a test-deployment and a prod deployment. If so we should probably mention each and what it entails. This info should be delegated to the builders' section.

### Builders 
I may need more information regarding what a builder does, however, this seems like the section of the library that would allow a user to expand on its functionality. If we decided to allow for custom builds they would have to fall into the test or prod deployment and secondly we have to define in this section exactly what parts are extensible and how we plan to make them so.
In my opinion, the whole power of the library will rest here. This will be where we define how we plan to leverage the library to build more complex software we should define very well how we plan to make this happen.
This section should exclusively be about getting the user to point us to the correct extensions it wants to make for its project. The build process should always look the same, save the test vs prod. The only thing that should vary is the inputs w build and that is what we should explain and deliver here.

### Baseline workgroup manager
The workgroup will be, ideally, one of the results of the builder defined. I would remove this section and move it to a more appropriately named one below

### Output
Create a new section in which we describe exactly what the output of the builder ill be, among others this will be the workgroup manager. But what else will be deployed? and what part of that deployment was baselined, as opposed to extended? Do the outputs vary depending on the inputs, I don't think they should but it's a question worth asking, especially if we want to make this extensible.

This is a huge PR and a great idea, I think it needs further polishing on the abstraction layer and we have to work on definitions and scope a bit more, but its a very good start.

Perseverance:
> If I understand the purpose of the PR correctly, the objective is to abstract the baseline contract deployment into a library. This library should also provide the ability to extend the baseline to accommodate for more complex projects.
> 
> The PR should, therefore, focus on how it will achieve this abstraction and extensibility, leaving behind how the process works in radish.
> 
> We should build this streamline of baseline independently of the radish use-case. I understand the difficulty of doing so, but building something generic from the base of a use-case will create technical debt, and we will have to come back and refactor this eventually.
> 
> Diving further into the PR I have several comments:
> 
> ### Logical Input Objects
> I think we should revisit the structure of this section like so:
> **Input:** Name of the input
> **Justification:** What purpose it serves, and why it is a baseline input
> **Input type:**
> 
> * _System Store_ - It is required for the initialization of the system but is, explicitly, not an extensible input
> * _User Input_ - The user determines the input and passes the parameters to the library on init
> * _Hybrid_ - A baselined input comes by default, and is required for the initialization, but can be extended by the user
>   **Structure:** A description of how the input will be fed to the library, and in what format
>   **Reference:** A mention of what is currently being used in radish34 to help us map the functionality and process.
> 
> An example would look like:
> 
> > **Input:** Contract Artifacts
> > **Justification:** Any baseline project will require the initialization of some basic contracts defined in this input (potentially even list the basic contracts? )
> > **Input type:** _Hybrid_ - A baselined input comes by default, and is required for the initialization, but can be extended by the user
> > **Structure:**  At a minimum, these should have the contracts bytecode, name and abi.
> > **Reference:** In Radish34 these are taken from the artifacts produced from the compilation.
> 
> @Perseverace I know your description contains most of these already but adding the structure and extra elements will help us form a better picture of how we will be abstracting the code. This will turn out to be a huge effort, and clarity will aid in the review.
> 
> ### Resolvers
> I am struggling to get a good grasp on what resolvers are. So please correct if I'm wrong The resolvers work as a middleware to translate the input given by the user to the input accepted by the library. While I think to see the usefulness of having this functionality, I think we are much better off hardcoding the type of the inputs instead of giving the user that flexibility. This can become very difficult to maintain with time, perhaps I'm missing something but this seems to be something the user could do without that doesn't contribute to the concept of “baselining”
> 
> ### Baseline Deployer
> This seems good overall, I would include an idea of how key management is going to be handled or explicitly leave that out of the scope of this library.
> We should also consider whether we want a baseline library for each part of the baseline or a single library that builds a baseline application from scratch. There are other parts too that require deployment and abstraction in a similar way we should discuss how we plan to bring, or not bring, them together.
> This suggests that we might be looking at two different kinds of deployments, a test-deployment and a prod deployment. If so we should probably mention each and what it entails. This info should be delegated to the builders' section.

Thanks for the comment. I agree with the comment about the Logical inputs, but I, unfortunately, disagree with several of the other comments. 
> ### Builders
> I may need more information regarding what a builder does, however, this seems like the section of the library that would allow a user to expand on its functionality. If we decided to allow for custom builds they would have to fall into the test or prod deployment and secondly we have to define in this section exactly what parts are extensible and how we plan to make them so.
> In my opinion, the whole power of the library will rest here. This will be where we define how we plan to leverage the library to build more complex software we should define very well how we plan to make this happen.
> This section should exclusively be about getting the user to point us to the correct extensions it wants to make for its project. The build process should always look the same, save the test vs prod. The only thing that should vary is the inputs w build and that is what we should explain and deliver here.
> 
> ### Baseline workgroup manager
> The workgroup will be, ideally, one of the results of the builder defined. I would remove this section and move it to a more appropriately named one below
> 
> ### Output
> Create a new section in which we describe exactly what the output of the builder ill be, among others this will be the workgroup manager. But what else will be deployed? and what part of that deployment was baselined, as opposed to extended? Do the outputs vary depending on the inputs, I don't think they should but it's a question worth asking, especially if we want to make this extensible.

The Builder is a helper class and the deployed does the main work. The builder allows for modular and partial deployment task definition. The resulting deployment task is picked by the deployed and the work is done there.
I do not see an advantage in returning a workgroup manager from the deployment procedure, especially as not all deployment procedures result in new workgroup contracts. So for these two I'd rather keep these two like they are now.

> This is a huge PR and a great idea, I think it needs further polishing on the abstraction layer and we have to work on definitions and scope a bit more, but its a very good start.
I feel that abstracting will be a lengthy process that will render this PR unmergeable. I'd much rather employ a 'divide-and-conquer' strategy and allow for further partial abstractions to be done later. I realise that we run the risk of technical debt, but I do not feel we are in a position for us to be able to understand our upcoming users enough to make the correct abstraction decisions. 

In short: Id much rather have this merged sooner rather than later, and address next abstractions in next PRs otherwise this will never move forward.

LucasRodriguez:
> The Builder is a helper class and the deployed does the main work. The builder allows for modular and partial deployment task definition. The resulting deployment task is picked by the deployed and the work is done there.
> I do not see an advantage in returning a workgroup manager from the deployment procedure, especially as not all deployment procedures result in new workgroup contracts. So for these two, I'd rather keep these two like they are now.

I agree they are two different things and the way I phrased it: 

> The workgroup will be, ideally, one of the results of the builder defined. I would remove this section and move it to a more appropriately named one below

Is wrong, It makes sense to keep it as its own component. 

> In short: Id much rather have this merged sooner rather than later, and address next abstractions in next PRs otherwise this will never move forward.

I'm a huge proponent of abstracting as much as we can, as soon as we can. When we base an abstraction on the only available reference implementation its not really an abstraction at all. 

It is also fine is this is not an abstraction but rather simply a restructuring of the deployment. I may have misunderstood the purpose of the PR.  I know it's semantics but where the effort should be directed depends on what the finality really is

> Thanks for the comment. I agree with the comment about the Logical inputs, but I, unfortunately, disagree with several of the other comments.

Happy to discuss further, either here or in a call

skarred14:
Great points raised about the scope and purpose of this PR @LucasRodriguez @Perseverance 
This PR is to separate out the deployment artifacts from the functionality, and ease of use for other reference implementations or customizations. A general representation for managing deployments is a good idea, and resolvers here look like they are purposed for the set available in Radish34. Builder is a facility/wrapper for setting up the scene - as we define further abstractions or define the builder and resolver in a uniform manner. 

As follow on consideration, It would be good to surface these abstractions to `baseline/` package manager, as a wrapper or extension of the core module. This would give better separation of what baseline core is vs. an applied form  (leveraging a specific set of contracts (from a "stdlib")) as a deployment wrapper using specific constructions: saving settings to a mongo db, OR compiling/deploying .sol type contracts (in the likely event of a shift to wasm), resolving the compilation of circuits written in a specific language/tool (starting with .zok or .circom, for example).

skarred14:
@Perseverance could you link the issues that this PR is related to, so that they can be closed with the merge of this pr

Perseverance:
> @Perseverance could you link the issues that this PR is related to, so that they can be closed with the merge of this pr

Done. Ready to be merged :)

PR #113, opened 2020-04-16
Combine integration test suites

bitwiseguy:
# Description

Currently there are two integration test suites: one uses Postman and another uses jest. In order to prevent redundancy and make it easier to compile an accurate test coverage report, we should combine those suites and remove redundant tests.

## How Has This Been Tested

`make test`

## Types of changes

- Added tests to `integration.test.js`
- Removed Postman integration tests.
- zkp unit test Postman suite no longer automatically run at integration level

## Checklist


radish34/__tests__/integration.test.js
```diff
@@ -158,12 +220,30 @@ describe('Buyer sends new RFP to supplier', () => {
...
+      expect(res.statusCode).toEqual(200);
+      expect(res.body.data.rfp._id).toEqual(rfpId);
+      expect(res.body.data.rfp.sku).toEqual(sku);
+    });
+
+    test('Supplier2 graphql query rfp() returns 200', async () => {
       const queryBody = `{ rfp(uuid: "${rfpId}") { _id, sku } } `
       // Wait for db to update
       let res;
       for (let retry = 0; retry < 10; retry++) {
```
skarred14, line 245: Is there a way these retries are made based on a parameter, so that user would know which test in the overall process is the wait/delay at?

radish34/__tests__/integration.test.js
```diff
@@ -271,17 +452,14 @@ describe('Buyer creates MSA, signs it, sends to supplier, supplier responds with
...
-      expect(res.body.data.createMSA.zkpPublicKeyOfSupplier).toEqual(supplierZkpPublicKey);
+      expect(res.body.data.createMSA.zkpPublicKeyOfBuyer).toEqual(buyer.zkpPublicKey);
+      expect(res.body.data.createMSA.zkpPublicKeyOfSupplier).toEqual(supplier2.zkpPublicKey);
       expect(res.body.data.createMSA.sku).toEqual('FAKE-SKU-123');
       expect(res.body.data.createMSA._id).not.toBeNull();
       expect(res.body.data.createMSA.commitments[0].commitment).toEqual(concatenateThenHash(
-        buyerZkpPublicKey,
-        supplierZkpPublicKey,
+        buyer.zkpPublicKey,
+        supplier2.zkpPublicKey,
```
skarred14, line 462: Would be good to add to the documentation/comments, about the scope/coverage of these tests, i.e. MSA for a specific org, etc.

radish34/__tests__/integration.test.js
```diff
@@ -100,7 +133,26 @@ describe('Buyer sends new RFP to supplier', () => {
             skuDescription: "Widget 200",
             description: "Widget order for SuperWidget",
             proposalDeadline: 1578065104,
-            recipients: [ { identity: "${supplierMessengerId}" } ]
+            recipients: [
```
skarred14, line 136: 👍

radish34/__tests__/integration.test.js
bitwiseguy: Are you asking if we can print logs while the test is running so the user knows which test is currently being run? Or are you suggesting that the number of retries is configurable and can be passed in by the user as a parameter?

radish34/__tests__/integration.test.js
bitwiseguy: Do you mean something like a test coverage report produced by jest combined with more descriptive `describe()` statements for the tests?

radish34/__tests__/integration.test.js
skarred14: i mean logs for a specific function so that user knows which test is currently the testing process is at

radish34/__tests__/integration.test.js
skarred14: For example, the tests here cover sending RFP to both suppliers, but MSA is only for one supplier. So just a minor comment to clarify that these MSA tests are for Supplier 2

Makefile
```diff
@@ -88,8 +88,6 @@ test:
 	@pushd ${radish34} && \
 	npm run await-stack && \
 	npm run test && \
-	npm run postman-test-zkp-api && \
```
breakpointer, line 91: I see you've deleted both postman test execution steps here but in the package.json only removed one of the tests. Are you combining them? How will they run as part of `make test` and/or `npm`?

radish34/__tests__/integration.test.js
```diff
@@ -43,6 +51,22 @@ afterAll(async () => {
   await nativeClient.close();
 });
 
+describe('Get organization settings from config files', () => {
```
breakpointer, line 54: Are you testing the test setup here? In the test? So meta... but might not be necessary 😛

Makefile
bitwiseguy: The one I deleted is the redundant integration-level test suite. The other one is a test suite specifically for the zkp service. I think all service-specific test suites should be run by `npm run test` from within the individual service. So that npm script should be moved into `zkp/package.json`. Do you agree?

radish34/__tests__/integration.test.js
bitwiseguy: Haha yes I see your point. I was thinking if one of these tests fails it would indicate to the user that something had gone wrong prior to the testing. Most likely, the deploy process failed to properly create these files. I can delete these tests though and just simply import the config files if that's preferred

radish34/__tests__/integration.test.js
bitwiseguy: Added some logs to long-running tests so user is more informed

radish34/__tests__/integration.test.js
bitwiseguy: Edited the test descriptions so that it is clear which parties are involved during each set of tests

skarred14:
all tests pass! appreciate the thoroughness in test case scenarios

Makefile
bitwiseguy: I moved the script into `zkp/package.json` to conform to structure in other service directories

radish34/__tests__/integration.test.js
breakpointer: Maybe this or some similar checks should go in the deploy app itself? Like, a post-deploy check? I know I’ve been bitten by this no-config-files bug.

PR #136, opened 2020-04-26
Comments in the code on instructions how to view Radish34 outside of localhost

danielnorkin:
# Description

Adding Comments in the code on instructions how to view Radish34 outside of localhost.

## Related Issue

N/A

## Motivation and Context

Setting up a demo in Azure and need to show the webapp outside of logging into the cm

## How Has This Been Tested

Yes. These are only comments

n/a

## Checklist

n/a


radish34/ui/src/components/UserSelection.js
```diff
@@ -29,6 +29,7 @@ const useStyles = makeStyles(() => ({
 const UserSelection = () => {
   const classes = useStyles();
   const [data, setData] = useState({ status: 504 });
+  // if you want radish34 to be seen outside of localhost (ie. Azure, AWS, etc.) replace const apiURL = window.localStorage.getItem('api') || 'localhost:8001/graphql' with const apiURL = window.localStorage.getItem('api') || 'public ip address:8001/graphql'
   const apiURL = window.localStorage.getItem('api') || 'localhost:8001/graphql';
```
skarred14, line 33: would it make sense to make this as an env var?

breakpointer:
We should probably figure out a way to push these out to config variables in the environment. I'd recommend that approach over editing the code directly. However, I do like the comments.

Issue #50, opened 2020-03-27
Reduce zkp integration test time


bitwiseguy:
## Overview

Currently the integration test suite (run by `make test`) takes several minutes to complete. All tests take less than a second, except for two tests that make calls to the zkp `/generateProof` endpoint.

### Reference

The generation of the proof is very time consuming. Your laptop has to whir away through lots of tedious elliptic curve operations for minutes. It’s nothing to do with the blockchain. Once the proof is generated (within the zokrates container), the act of sending the proof to the shield contract as a blockchain transaction is very quick (especially when using ganache).

> Note: the periodic "socket hangup" issue that produces test failures is caused by the long process time of POST /generateProof. This should be solved by replacing that http REST call with bulljs queues. That work is currently on the "development" branch and will be eventually merged into "master"

### Potential solution

What we could do is create a set of lazy dummy arithmetic circuits (i.e. dummy files for createMSA.zok and createPO.zok where the body of the function just contains a dummy line of useless code). This would make testing much faster. Developers would just have to be cautious when editing any of the backend zkp code (in case they break the "proper" .zok files).

> Note: createMSA.zok and createPO.zok  are already quite ‘simplified’  to reduce proving time, because the merkle tree being used only has 16 leaves (which takes a minute or two to generate the proof) , rather than 4 billion (which takes around 5 minutes). 

We could use the "dummy" circuits for integration testing and use the "real" circuits for unit testing of the `zkp` service.

kthomas:
@skarred14 was this addressed by the Makefile effort?

skarred14:
@kthomas @bitwiseguy 

this is a WIP but it has the dummy circuits, just need to build the integrations or switches for these dummy circuits.
https://github.com/ethereum-oasis/baseline/tree/dummy-circuits

PR #149, opened 2020-05-14
feat: dummy circuits

skarred14:
# Description

Reduce integration test timing for faster execution of tests. This work is to simplify the circuits which are used for creation of an MSA and a PO.

## Related Issue

#50 

## Motivation and Context

Reduces the test time from around 3 to 4 minutes to under 30 seconds. If a "mode" argument is not passed to `make start`, then everything defaults to using the original `createMSA` and `createPO` circuits.

## How Has This Been Tested

`make && make start && make test`
`make && make start mode=0 && make test`
`make && make start mode=1 && make test`

All command sequences above pass the integration test suite.

## Types of changes

## Checklist

- [X] My change requires a change to the documentation.


radish34/ops/setup_circuits.sh
```diff
@@ -9,18 +9,33 @@ NC='\033[0m'
 docker-compose up --no-recreate -d zkp
 ./ops/await_stack.sh
 
-if [ ! -d ./zkp/output/createMSA ]; then
+if [ "$1" = "" ]; then
```
breakpointer, line 12: I like this change a lot. However, I think the way it's implemented with circuit selector flags could be better. It feels like the option should be in a config file or somewhere else. Or perhaps, instead of looking for an explicit flag it looks for an environment? Like, `ENV == "CI"` then use dummy?
Or maybe there's a way to swap out the files at the volume level such that the dummy circuits are just aliased or drop-in replacements for the real circuits? Something like you might do with sym links...

I'm not sure I'm explaining this well but there's something about the new `ZKP_MODE` in all the services in the compose files that just seems a bit out of place.

skarred14: To make sure there is a consistent use of a test vs dev mode, one way that could be handled is to store an e.nv file at the root of the repo which can be used across `Dockerfile`/docker-compose environment parameters and `.sh` files.

In terms of symlinks, i believe that could be better handled at the baseline/ level to provide for choosing between test vs real toggle for circuit-set selection. Or should we handle this as part of this pr: create sym-links as additional lines in `ops/build.sh`? Thoughts? @breakpointer @bitwiseguy

bitwiseguy: I spoke with @kthomas yesterday about dynamically creating docker-compose files based on our different environment types (ci vs. dev vs. prod). An example of that is [here](https://github.com/connext/indra/blob/staging/ops/start-dev.sh). I plan on starting a new branch to contain that work. Could we address the circuit selection that way in a separate PR by injecting a "circuit-selection" environment variable to the containers based on the environment type? Or do you all have any thoughts on the dynamically generated docker-compose files in general?

skarred14: agreed, this is a much refined approach to handle custom selections of circuit sets and aligning with the different environments. I concur on addressing circuit-selector work to a separate PR. We can track it as its own issue. Does the approach sound ok to you @breakpointer

kthomas: I like the change also. It makes sense to parameterize which circuit(s) are under test or otherwise required.

Also agree that environment variables should not "leak" out of the service they are intended to be set for.

breakpointer: Sure 👍

PR #151, opened 2020-05-19
Import Solidity contracts

bitwiseguy:
# Description

- Changes the `radish34/contracts` compiler from `0x/sol-compiler` to Truffle
- Now we are feeding the pre-compiled `radish34/contracts/artifacts` JSON files (containing the abi) to `merkle-tree` service instead of making that service re-compile a second time

## Motivation and Context

These changes allow us to properly use Solidity contract imports instead of having to maintain copies of those contracts in this repo.

Previously, the `merkle-tree` service re-compiled all of the contracts on its own using `solc` compiler because it needed the build artifacts in `solc`'s format. Compiling the same contracts twice was inefficient. 

## How Has This Been Tested

From `baseline/radish34/contracts`:
```
npm run build
docker-compose up -d ganache
npm run test
```

From `baseline`:
```
make && make start && make test
```

## Types of changes

## Checklist


radish34/docker-compose.yml
```diff
@@ -447,13 +447,13 @@ services:
...
       - ./config/merkle-tree:/app/config # mount point might be different if configuring from another application
-      - ./contracts/contracts:/app/contracts:consistent # required if deploying/compiling contracts from within this service (if CONTRACT_ORIGIN = 'default' or 'compile')
+      - ./contracts/artifacts:/app/build/contracts:consistent # required if deploying/compiling contracts from within this service (if CONTRACT_ORIGIN = 'default' or 'compile')
     ports:
       - 9000:80
     environment:
       BLOCKCHAIN_HOST: ws://ganache
       BLOCKCHAIN_PORT: 8545
-      CONTRACT_ORIGIN: 'compile' # Where to find the contractInstances?
+      CONTRACT_ORIGIN: 'default' # Where to find the contractInstances?
```
breakpointer, line 456: I'm okay with the change but a little concerned by this statement in the Timber repo. "Importantly, Timber can only interpret smart contract interface json files which have been compiled with solc." So while the test pass (I'm assuming) there might be some unexpected behavior from Timber. https://github.com/EYBlockchain/timber#contract_origin 
Have you looked into the impact switching to truffle will have on Timber? Or asking them directly since we have some connections there?

bitwiseguy: I found that the sentence you quoted from Timber is not completely accurate. What Timber does need is the build artifact JSON files to be formatted as described in this function: https://github.com/EYBlockchain/timber/blob/master/merkle-tree/src/compile.js#L62-L75. That is actually the same format that Truffle uses but is different than the default output format of `0x/sol-compiler`. That is why we had to force the `merkle-tree` service to recompile the files itself when using `0x/sol-compiler` within our `radish34/contracts` build.

`0x/sol-compiler` outputs JSON in this format:
```
{
  compiledOutput:
    abi: "...",
    evm:
      bytecode: "0x..."
}
```

`truffle compile` outputs JSON in this format:
```
{
  abi: "...",
  bytecode: "0x...."
}
```

bitwiseguy: I have looked into the impact of switching to truffle only as far as running our tests and monitoring the timber logs to confirm there are no errors reported. I have not reached out to any of the EY research team to ask them about any unintended consequences

radish34/contracts/package.json
```diff
@@ -5,19 +5,19 @@
   "main": "src/utils.js",
   "scripts": {
     "test": "jest --runInBand --forceExit --detectOpenHandles",
-    "build": "sol-compiler"
+    "build": "truffle compile"
```
Perseverance, line 8: I'd like to give the cheeky suggestion that `etherlime` will do the same thing and not add web3.js to our dependencies ;)

bitwiseguy: @Perseverance I'm not very familiar with `etherlime`'s capabilities. Are you suggesting to use it to compile the contracts rather than using `truffle`?

radish34/deploy/src/baseline-administrator-lib/resolvers/contract-resolvers/radish-configpath-resolver.js
```diff
@@ -59,55 +59,52 @@ class RadishConfigpathContractsResolver {
...
-			contractName,
-			abi,
-			bytecode
-		}
-	}
+  async resolve(name) {
+    assert(name, 'No name supplied to the resolver resolve method');
+    const contractPath = path.resolve(this.paths[name]);
+    const contractJSON = require(contractPath);
+    const { contractName, abi, bytecode } = contractJSON;
```
Perseverance, line 78: If the bytecode is comming directly from truffle compile, it might come with its libraries linked already (not sure how they decide what address to use). Please verify, if you have not, that the deployment of contracts with libraries work correctly.

bitwiseguy: As far as I know the library linking is working same as before: the contract bytecode is formatted like this - `0x<contract_source_code_in_hex>__<libraryName>__<library_source_code_in_hex>`

The main difference in deployer is we no longer need to utilize the [_replaceLibraryReferenceToTruffleStyle() function](https://github.com/ethereum-oasis/baseline/blob/master/radish34/deploy/src/baseline-administrator-lib/resolvers/contract-resolvers/radish-configpath-resolver.js#L28) because `truffle` links the libraries within the bytecode in that format by default. Previously, `0x/sol-compiler` output bytecode that linked libraries using the hash of the library instead of the `libraryName` so you were using the `_replaceLibraryReferenceToTruffleStyle` function to convert the hash to the `libraryName`.

I have run the deployer via `make && make start && make test` and the deployer seems to work as well as the current `master` branch deployment. I'm not clear what you mean by "not sure how they decide what [library] address to use". I believe the libraries are part of the newly deployed contract bytecode, not references to previously deployed contracts with their own addresses.

radish34/contracts/package.json
skarred14: should we consider a switch or an option to be added to use etherlime instead? Does etherlime use truffle as well? @Perseverance

radish34/docker-compose.yml
breakpointer: Okay, it might be nice to tell them this. A PR to their project with your findings perhaps?

radish34/deploy/src/baseline-administrator-lib/resolvers/contract-resolvers/radish-configpath-resolver.js
Perseverance: Hey Sam, Since it is working I am good.

radish34/docker-compose.yml
bitwiseguy: @breakpointer I created this PR to update their README: https://github.com/EYBlockchain/timber/pull/12

PR #152, opened 2020-05-20
Logger

### LINKED ISSUE #101: Install logging framework and remove all `console.log` from code
## Overview

As a proper system, so I can be compliant with log level support, install a logging framework (like Winston) and remove all usages of the generic console.log. Replace them with appropriate 'info', 'warn', 'error', 'debug' level logging calls.

### Assumptions

- We will use the same logging format across all services
- The logs for HTTP requests vs system calls should be handled
- Log tracing is not required at this point but would be nice
- Log rotation/truncation is out of scope
- Logs should in their own folder and added to gitignore

### Acceptance

Minimum effort is to 
- install logger framework
- remove console.logs and replace the log call with the appropriate level 'info' 'warn' 'debug' etc...
- adjust environment configs to the appropriate log level (e.i. 'debug' in development, 'info' in production) 
- update documentation on how to view access the log files
- adjust logging format for all services such that they are the same.

Ideal logging format is 
[UTC timestamp] [service name] [host] - <logging message>

myronrotter:
# Description

Closes #101 
- Add customized winston logger for HTTP requests and system calls
- Replace generic `console.log` with uniform logging level calls

- [x] Install winston logging framework; Add uniform format; Integrate into build process as parent/base image
- [x] Use the same logging format across all services: `[UTC timestamp] [log level] [service name] [host]: ...`
- [x] Logging for HTTP requests and system calls
- [x] Update documentation

### What can be improved

- Adjust/add environment configs for logging level (currently, logger adjusts logging level according to `process.env` distinguishing between 'debug' in development, 'info' in production)
- Aggregate logs (currently in docker container `app/logs/`)

## Related Issue

Refer to #101 

## How Has This Been Tested

`make && make start && make test`
and check
```
docker-compose logs -f ui
docker-compose logs -f zkp
docker-compose logs -f api-buyer
docker-compose logs -f messenger-buyer
docker-compose logs -f api-supplier1
docker-compose logs -f messenger-supplier1
docker-compose logs -f api-supplier2
docker-compose logs -f messenger-supplier2
```

## Types of changes

## Checklist


radish34/docker-compose.yml
```diff
@@ -2,6 +2,12 @@ version: '3.5'
 
 services:
 
+  logger:
```
breakpointer, line 5: Does this service aggregate the logs too? I'm unclear why it's both a base docker image and also a service.

radish34/logger/dev.Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:11.15
```
breakpointer, line 1: Is this based on alpine? If not, can you use the alpine based node image instead?

radish34/messenger/Dockerfile
```diff
@@ -1,15 +1,25 @@
-FROM node:12-alpine
+FROM radish34_logger as logger
+FROM node:12.16
 
-# Create app directory
-WORKDIR /app
+ENV FORCE_COLOR=1
+
+RUN mkdir /logger
+WORKDIR /logger
```
breakpointer, line 7: It seem like there are some redundant statements in this docker file. Do you need to do the two above statements in this file since they're already part what you do in the base image `radish34_logger`?

radish34/zkp/src/utils/jsonifyVk.js
```diff
@@ -50,7 +50,6 @@ const keyExtractor = solFilePath => {
  */
 export const jsonifyVk = async solFilePath => {
   const vkJson = await keyExtractor(solFilePath);
-  // console.log('vkJson:', vkJson);
```
breakpointer, line 53: `logger.debug` here instead of removing.

radish34/messenger/config/development.js
```diff
@@ -5,9 +5,6 @@ module.exports = {
     bufferMaxEntries: 8,
     firstConnectRetryDelaySecs: 5,
   },
-  logging: {
```
breakpointer, line 8: So where do we set the log level if not here?

radish34/logger/dev.Dockerfile
myronrotter: AFAIK it is not based on alpine. I sticked with `node:11.15` for the logger image, since most of the services used `node:11.15` except of `deploy` and `messenger` services (which use 12ish and 13ish versions). Since the services such as zkp rely on node-gyp for dependencies switching to alpine may cause some issues and/or need some workaround.

radish34/messenger/config/development.js
myronrotter: Am I right saying this configs only hold for the messenger services? So what about the other services? A common way to handle this would be nice, so any suggestions?

radish34/api/src/services/merkle-tree.js
```diff
@@ -201,31 +194,17 @@ export const checkRoot = (commitment, commitmentIndex, siblingPath, root) => {
     .padStart(TREE_HEIGHT, '0') // pad to correct length
     .split(''); // convert to array for easier iterability
 
-  // console.log(`commitment:`, commitment);
```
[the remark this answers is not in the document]
myronrotter: Sure, added the `logger.debug` respectively.

radish34/api/src/utils/crypto/hashes/mimc/mimc.js
```diff
@@ -80,28 +80,24 @@ export const mimcHash = (...preimage) => {
     .toString(16)
     .padStart(64, '0')}`;
 
-  console.log(hash);
```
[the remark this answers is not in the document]
myronrotter: Done.

radish34/api/src/utils/ethers.js
```diff
@@ -213,9 +214,6 @@ const getEventValuesFromTxReceipt = (eventName, contract, txReceipt) => {
       let { values } = logDescription;
       values = removeNumericKeys(values); // values contains duplicate numeric keys for each event parameter.
       values = parseBigNumbers(values); // convert uints (returned as BigNumber) to numbers.
-
-      // console.log(`\n\nExtracted these values relating to ${eventName}:`);
```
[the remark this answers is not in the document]
myronrotter: Done.

radish34/zkp/src/utils/jsonifyVk.js
myronrotter: Done.

radish34/messenger/Dockerfile
myronrotter: Indeed, those statements look kind of redundant. But I would prefer these statements instead of copying everything (including `src` files and `node_modules`) from the `radish34_logger` image with `COPY --from=logger /logger/ /logger/`.

radish34/docker-compose.yml
myronrotter: No, this is not a service that aggregates the logs. At first I tried to come up with a solution for log aggregation and those lines are some leftovers. I think this can be addressed in another issue/PR.

radish34/messenger/config/development.js
kthomas: Agree this needs to live in a common package. We have an integration branch that stands up these shared artifacts called `init-core` @myronrotter. Let's get this merged into master and then we can figure out how to best expose it in shared utils package of some kind.

breakpointer:
Okay great! Glad we’ve got approval. Now it appears there are some merge conflicts. @myronrotter could you take one last pass at this to correct? Thanks again for your efforts.

PR #243, opened 2020-09-29
Adding Oasis Standard template for Baseline CORE specification.

Consianimis:
# Description
We were provided with OASIS templates that need to be used for the drafting of the Baseline Specifications.
This should be used by editors who have signed the iCLA/eCLA.


skarred14 (dismissed):

chaals (approved):
Looks like a reasonable starting point...

humbitious (approved):
Reviewed structure with Anais. Looks right. Some things may change, but this is good for starting.

bitwiseguy (approved):

kthomas (approved):

PR #274, opened 2020-12-11
radish34 - pinning geth to a version that supports whisper

vivekrao1985:
# Description

What's in this PR -
- Pinning radish34 to geth version ```1.9.20```, the last version that supports whisper. See the release notes for [1.9.21](https://github.com/ethereum/go-ethereum/releases/tag/v1.9.21).
- Adding logger to ```docker-compose.yml``` in order to make the documentation consistent with environment creation.
- Update jest to ```24.9.0```.
- Added better logging in the integration tests.

## Related Issue

I'm open to discussing the above points and opening a ticket if deemed necessary.

## Motivation and Context

Why is this change required?
It's required to help first time devs get set up with radish34 by following the documentation.

What problem does it solve?
This will help people spin up radish34 locally just by following the documentation.

## How Has This Been Tested

Before:
- Running ```make npm-install``` and then running ```docker-compose build``` as per the documentation gives the following error -
```
ERROR: Service 'zkp' failed to build : pull access denied for radish34_logger, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
- Running ```make test``` leads to a lot of failing integration tests due to the absence of whisper from the latest geth version.

After:
- Running the same commands in the same sequence creates the environment successfully.
- Integration tests pass.

## Types of changes

## Checklist

- [ ] My code follows the code style of this project.


bitwiseguy (approved):

bitwiseguy (dismissed):

bitwiseguy (approved):

bitwiseguy (approved):

bitwiseguy (dismissed):

tkstanczak (approved):

examples/radish34/geth-env/bootnode/Dockerfile
```diff
@@ -6,11 +6,16 @@ RUN apt-get update \
...
+
+# Pinning geth to version 1.9.20 as all successive geth versions dont support whisper anymore
+# Refer to the release notes for 1.9.21 - https://github.com/ethereum/go-ethereum/releases/tag/v1.9.21
+RUN wget https://gethstore.blob.core.windows.net/builds/geth-linux-amd64-1.9.20-979fc968.tar.gz
+RUN tar -xzvf geth-linux-amd64-1.9.20-979fc968.tar.gz
+RUN chmod +x ./geth-linux-amd64-1.9.20-979fc968/geth
+RUN mv ./geth-linux-amd64-1.9.20-979fc968/geth /usr/local/bin/
+RUN rm -rf ./geth-linux-amd64-1.9.20-979fc968
 
 CMD exec bootnode -nodekeyhex $nodekeyhex
```
BrianCraig, line 20:
```suggestion
CMD exec geth -nodekeyhex $nodekeyhex
```

BrianCraig (changes requested):

BrianCraig:
Hi! 
What is the purpose of the logger container? It starts and it stops automatically because the start command is `npm run build`.

Thanks, Brian

BrianCraig:
I noticed that after using this commit there is a new volume from geth that utilizes 80~GB, 
Screenshot from 2021-01-15 21-50-43
I may think this is the production eth chain, is this correct ?

kthomas:
Merging this PR is a bit dangerous in that doing so implies that whisper is a supported or suggested way to implement the protocol. This could mislead a lot of folks potentially. Good looking PR otherwise :)

vivekrao1985:
> Merging this PR is a bit dangerous in that doing so implies that whisper is a supported or suggested way to implement the protocol. This could mislead a lot of folks potentially. Good looking PR otherwise :)

That's a great point. The problem is people already refer to this as an example to implement the protocol. So they're spending time looking at something wrong that isn't even working out of the box. Would it be better to remove radish34 entirely?

kthomas:
> > Merging this PR is a bit dangerous in that doing so implies that whisper is a supported or suggested way to implement the protocol. This could mislead a lot of folks potentially. Good looking PR otherwise :)
> 
> That's a great point. The problem is people already refer to this as an example to implement the protocol. So they're spending time looking at something wrong that isn't even working out of the box. Would it be better to remove radish34 entirely?

Radish34 was great from a POC/use-case perspective but it is obsolete in the context of a protocol reference implementation (it was never actually a reference implementation). It also has a fair amount of technical debt that would need to be overcome, as you noted.

At a very minimum we should update the README to indicate/warn users that it has been deprecated and still here for historical purposes. Could also add a link there to [BRI-1](https://github.com/ethereum-oasis/baseline/tree/master/examples/bri-1/base-example) or just the more general [overview of reference implementations](https://docs.baseline-protocol.org/bri/overview-of-reference-implementations).

kthomas:
@vivekrao1985 you should update the README as described above and then I think we can merge this PR since it is incrementally better thanks to your effort. It would also have a deprecation warning and instructions for users.

A decision to remove Radish34 entirely would be a decision for the TSC to consider I think. It is probably worth discussing though to ensure the community invests time in the right places. I will bring your concern up at the next meeting (which happens to be today).

kthomas (changes requested):
Add deprecation warning and instructions to README

breakpointer (dismissed):

examples/radish34/geth-env/bootnode/Dockerfile
vivekrao1985: I think the purpose of the bootnode is for new nodes to be able to connect to the private network and find the rest of the cluster. The bootnode command brings up a stripped down version of geth. Here's a [doc](https://geth.ethereum.org/getting-started/private-net) describing the whole process.

vivekrao1985:
@kthomas Yup, makes sense. I updated the README.

kthomas (approved):

tkstanczak (approved):

PR #282, opened 2020-12-15
Update core persistence and privacy types

kthomas:
# Description

Update the core `types` package to support more robust modeling of multiparty baseline workgroups and workflows.

This PR sets the stage for a follow-on PR which will add an initial in-memory provider implementation to the core `persistence` package.

## Related Issue

- Microsoft Excel persistence package hackathon challenge -- this PR will make it more clear how a plugin can be defined to work with the core persistence package.

## Motivation and Context

The baseline standards initiative needs more clearly articulated data models. Developers need to be able to see how, for example, an Excel persistence provider could be created.

### DIFF
--- a/core/types/src/persistence.ts
+++ b/core/types/src/persistence.ts
@@ -1,38 +1,39 @@
-export type Commitment = {
-  // currentHash = H(Underlying data/doc + salt)
-  salt: string;
-  // Values needed for verification (Shield.verifyAndPush inputs)
-  value: string;
-  proof: number[];
-  publicInputs: string[];
-  signatures: object; // Allows mapping from participant address -> signature
-  metadata: object;
-  sender: string; // Address of participant who created the commitment
-};
+import { Circuit, Commitment } from './privacy';
 
 export type Participant = {
-  address: string; // Can use this to lookup messagingEndpoint, etc.
-  metadata: object;
+  address: string; // public address of the participant
+  metadata: object; // arbitrary metadata
+  url?: string; // optional url
 }
 
-export type State = {
+export type Store = {
+  identifier?: string; // optional system of record identifier, i.e. document name, UUID or primary key of relational record
+  metadata?: object; // arbitrary metadata
+  provider: string; // i.e., the system of record persistence provider; see implementations within `core/persistence/providers`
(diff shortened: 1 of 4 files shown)


core/types/src/privacy.ts
```diff
@@ -0,0 +1,21 @@
...
+  metadata?: object;
+}
+
+export type Commitment = {
+  metadata: object; // arbitrary data
+  location: number;
+  salt: string; // salt such that currentHash = H(data + salt)
+  proof: number[];
+  publicInputs: string[];
+  sender: string; // address of participant who created the commitment
```
bitwiseguy, line 18: Which address is this meant to be? The zkp public key? Would be good to update the comment to clarify

core/types/src/privacy.ts
```diff
@@ -0,0 +1,21 @@
...
+  signatures: object; // allows mapping from participant address -> signature
+  value: string; // values needed for verification (Shield.verifyAndPush inputs)
```
bitwiseguy, line 20: The inputs for `Shield.verifyAndPush()` include:
- proof
- publicInputs
- value/commitment
Can we update the comment here to make it clear `value === commitment`?

core/types/src/persistence.ts
```diff
@@ -1,38 +1,36 @@
...
-  signatures: object; // Allows mapping from participant address -> signature
-  metadata: object;
-  sender: string; // Address of participant who created the commitment
-};
+import { Circuit, Commitment } from "./privacy";
 
 export type Participant = {
-  address: string; // Can use this to lookup messagingEndpoint, etc.
-  metadata: object;
+  address: string; // public address of the participant
```
bitwiseguy, line 4: Which address is this? Ethereum public key?

kthomas: Yes.

core/types/src/privacy.ts
kthomas: This matches the `Participant` address.

bitwiseguy (dismissed):

kthomas:
@Consianimis 👀

bitwiseguy (approved):

stefschmidt (approved):

marten42 (approved):

PR #302, opened 2021-01-18
refactor bri-1 tests

BrianCraig:
# Description

Small changes of the BRI-1 Tests
 -  Changed passing of data between `e2e.spec.ts` and `shared.ts`, from context to parameters
 -  Added await for initialization of Alice and Bob in tests (prevents race condition)
 -  Simplified retry of services
 -  Added explicit waiting for zkSNARK circuits sync (prevents race condition)

## Related Issue

No related issue

## Motivation and Context

Because it helps with the tests code quality and prevents some race conditions test fails

## How Has This Been Tested

Tested with debugger and checked for the usefulness of the explicit waiting for solving race conditions

## Types of changes

## Checklist

### DIFF
--- a/examples/bri-1/base-example/test/shared.ts
+++ b/examples/bri-1/base-example/test/shared.ts
@@ -1,6 +1,8 @@
 import { assert } from 'chai';
 import { ParticipantStack } from '../src';
 
+const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms));
+
 export const shouldBehaveLikeAWorkgroupOrganization = (getApp: () => ParticipantStack) => () => {
   describe(`organization details`, () => {
     let org;
@@ -140,11 +142,15 @@ export const shouldBehaveLikeAWorkgroupOrganization = (getApp: () => Participant
 
     describe('workflow privacy', () => {
       let circuit;
-  
+
       describe('zkSNARK circuits', () => {
         describe('synchronization', () => {
           before(async () => {
-            circuit = await getApp().getBaselineCircuit();
+            // We need to Wait fo the Sync
+            while (getApp().getBaselineCircuit() === undefined) {
+              await sleep(10000);
+            }
+            circuit = getApp().getBaselineCircuit();
             assert(circuit, 'setup artifacts should not be null');
           });
 
@@ -330,7 +336,7 @@ export const shouldBehaveLikeAnInitialWorkgroupOrganization = (getApp: () => Par
 
           it('should track the workgroup shield in an off-chain merkle tree database', async () => {
             // @ts-ignore
-            const trackedShieldContracts = await getApp().baseline.getTracked(); 
+            const trackedShieldContracts = await getApp().baseline.getTracked();
             assert(trackedShieldContracts.indexOf(shield.address.toLowerCase()) !== -1, 'workgroup shield contract should have been tracked');
           });
 
--- a/examples/bri-1/base-example/test/shared.ts
+++ b/examples/bri-1/base-example/test/shared.ts
@@ -1,11 +1,12 @@
 import { assert } from 'chai';
+import { ParticipantStack } from '../src';
 
-export const shouldBehaveLikeAWorkgroupOrganization = function () {
+export const shouldBehaveLikeAWorkgroupOrganization = (getApp: () => ParticipantStack) => () => {
   describe(`organization details`, () => {
     let org;
 
     before(async () => {
-      org = this.ctx.app.getOrganization();
+      org = getApp().getOrganization();
       assert(org, 'org should not be null');
     });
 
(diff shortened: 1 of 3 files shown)


examples/bri-1/base-example/test/shared.ts
```diff
@@ -138,11 +142,15 @@ export const shouldBehaveLikeAWorkgroupOrganization = function () {
...
       let circuit;
-  
+
       describe('zkSNARK circuits', () => {
         describe('synchronization', () => {
           before(async () => {
-            circuit = await this.ctx.app.getBaselineCircuit();
+            // We need to Wait fo the Sync
+            while (getApp().getBaselineCircuit() === undefined) {
+              await sleep(10000);
```
BrianCraig, line 151: Since the last draft this is the part that i changed, i noticed that the tests we're not passing because it wasn't waiting for the Sync to happen. I added debugger notices and checked that usually waits at least 10s

kthomas:
These changes-- do they actually work? I am hard pressed to actually get the test suite to fail and I know others in the community (@Meuko, @bitwiseguy, @skarred14 for example) have reported stability as well.

I'm nowhere near being convinced that these changes are actually needed and would like to understand what you're seeing on your end in a bit more detail.

examples/bri-1/base-example/test/shared.ts
kthomas: At this point in the test suite, we have already awaited the creation/delivery of the circuit. You're right in that the `await` is not necessary. But this just adds +10 seconds to the test suite without fixing any race conditions.

BrianCraig: Kyle, at my best understanding, comparing the base commit against this there is no change on the order of the actions on the test, but probably for the use of the function *tryTimes* it executes more early into this specific action, when the baseline circuit is not present in the app.
I added a debugger breakpoint and checked that usually with this changes the *baselineCircuit* is not yet populated on this specific commit, and sooner or later when this test evolves it should get to the same problem if it is not awaited to populate.

The exact times (10s) is a suggestion and can be changed to smaller times (5ms for example)

BrianCraig:
> These changes-- do they actually work? I am hard pressed to actually get the test suite to fail and I know others in the community (@Meuko, @bitwiseguy, @skarred14 for example) have reported stability as well.
> 
> I'm nowhere near being convinced that these changes are actually needed and would like to understand what you're seeing on your end in a bit more detail.

Hi! I had problems with my tests and resulted to be linked to my DNS provider for some reason, after changed to Google DNS i almost every time pass the full test suite in this commit and the base commit

kthomas:
> > These changes-- do they actually work? I am hard pressed to actually get the test suite to fail and I know others in the community (@Meuko, @bitwiseguy, @skarred14 for example) have reported stability as well.
> > I'm nowhere near being convinced that these changes are actually needed and would like to understand what you're seeing on your end in a bit more detail.
> 
> Hi! I had problems with my tests and resulted to be linked to my DNS provider for some reason, after changed to Google DNS i almost every time pass the full test suite in this commit and the base commit

We noticed some DNS issues were happening from Europe as well... do you know what the issue actually was?

I motion we close this PR :)

cc: @Meuko @ognjenkurtic

BrianCraig:
> > > These changes-- do they actually work? I am hard pressed to actually get the test suite to fail and I know others in the community (@Meuko, @bitwiseguy, @skarred14 for example) have reported stability as well.
> > > I'm nowhere near being convinced that these changes are actually needed and would like to understand what you're seeing on your end in a bit more detail.
> > 
> > 
> > Hi! I had problems with my tests and resulted to be linked to my DNS provider for some reason, after changed to Google DNS i almost every time pass the full test suite in this commit and the base commit
> 
> We noticed some DNS issues were happening from Europe as well... do you know what the issue actually was?
> 
> I motion we close this PR :)
> 
> cc: @Meuko @ognjenkurtic

Sincerely, i could get analytics about the times for each action and why there are multiple race conditions present on this test, and why you should avoid them in testing (sooner or later one microservice doesn't take the usual time, and could take 10 times more, and one action is called before another expecting otherwise, and the tests fail), but i assume you already know the importance of it. It is not my fault if for you this PR doesn't add anything to the repository, i'm sure this is something valuable and i invested time because i think it's important.

Sincerely, Brian

BrianCraig:
I'm sorry to continue this thread, but i understand why the need to wait for the syncing arises, and is because there is a substantial change with the `tryTimes` function, that intents to simplify the constant call for a promise, awaiting for the first successful response.
The original code that intents to simplify is this cases: 
```js
    let interval;
    const promises = [] as any;
    promises.push(new Promise<void>((resolve, reject) => {
      interval = setInterval(async () => {
        const vaults = await Vault.clientFactory(
          tkn!,
          this.baselineConfig.vaultApiScheme!,
          this.baselineConfig.vaultApiHost!,
        ).fetchVaults({});
        if (vaults && vaults.length > 0) {
          vault = vaults[0];
          resolve();
        }
      }, 2500);
    }));

    await Promise.all(promises);
    clearInterval(interval);
    interval = null;

    return vault;
```

This code for example i changed it for this
```js
    return await tryTimes(async () => {
      const vaults = await Vault.clientFactory(
        tkn!,
        this.baselineConfig.vaultApiScheme!,
        this.baselineConfig.vaultApiHost!,
      ).fetchVaults({});
      if (vaults && vaults.length > 0) {
        return vaults[0];
      }
      throw new Error();
    });

```

Because there were multiple usages of this interval type of code, i abstracted the idea of the code, to retry to get some information, or get a response from a server, and created the following function
 
```js
const tryTimes = async <T>(prom: () => Promise<T>, times: number = 80, wait: number = 9400): Promise<T> => {
  const errors : any[] = [];
  for (let index = 0; index < times; index++) {
    try {
      return await prom()
    } catch (err) { 
      errors.push(err);
    }
    await sleep(wait);
  }
  const error = new TryError("Unfulfilled promises");
  error.promiseErrors = errors;
  throw error;
}
```

which in fact, this change has the unique difference from the last one that *the promise expected to deliver a response, its called immediately* in difference of the setInterval, that waits 2.5s (in this case) before querying the promise. So the new race conditions come from here, leveraging unnescesary waiting times for this types of functions.

Thats is why, sooner or later, the tests would fail.

Sorry if this is too long, you can close the PR, but i'm confideent about my point of view.

Thanks

kthomas:
> 
> Thats is why, sooner or later, the tests would fail.
> 
> Sorry if this is too long, you can close the PR, but i'm confideent about my point of view.
> 
> Thanks

Nice work on refactoring the `require*` pattern to be more DRY... thanks for this and I understand it. It's much nicer.

I will try to run this locally. Which tests are still prone to racing without these changes? Thanks!

Meuko (commented):
Looks nice, nice changes! Thanks @BrianCraig !! Let me know if there's any conclusion to the comment exchange in shared.ts' review

PR #350, opened 2021-05-18
[bri-2] Zkp-mgr signature circuit generation

bitwiseguy:
# Description

The updates in this PR allow the `zkp-mgr` service to automatically generate zero knowledge circuits that can check for any number of eddsa signatures.

## Details

- update `gnark` version to v0.4.0
- fix encoding of inputs into witness for `gnark` to process
- add flexibility for service to handle hex values sent via body of request with or without `0x` prefix
- change name of "consistency" to "signature" circuit
- fix makefile bug that caused commands to fail if .env files are not present


Meuko (approved):

examples/bri-2/zkp-mgr/tests/example-circuits/zkp-mgr.postman_collection.json
```diff
@@ -0,0 +1,641 @@
...
+						"zkcircuits",
+						"{{circuit_id}}",
+						"verifier"
+					]
+				}
+			},
+			"response": []
+		},
+		{
+			"name": "all zkcircuits",
```
GrandGarcon, line 171: pretty descriptive api queries . in case can there be another endpoint query type for  returning the circuit result based on the circuit type / id ( consistency / cubic in this case ) ?

kthomas (approved):

PR #357, opened 2021-06-16
Update maintainers and codeowners

bitwiseguy:
# Description

Remove Brian Chamberlain and George Spasov from the maintainers list because they have decided to reduce their role in the baseline project to focus on other projects in the blockchain space.


Meuko (approved):
Looking good Sam, thanks!

skarred14 (approved):

PR #299, opened 2021-01-12
Add provide privacy to BRI-1

kthomas:
# Description

Add [Provide Privacy](https://docs.provide.services/privacy) to BRI-1.

Provide Privacy is a golang-based microservice that delivers high performance by way of [gnark](https://github.com/ConsenSys/gnark). This PR makes use of the `zkSnarkCircuitProviderServiceProvide` [privacy provider](https://github.com/ethereum-oasis/baseline/blob/master/core/privacy/src/zkp/provide.ts).

In addition, counterparties within a workgroup receive workgroup state, such as circuit artifacts (i.e., r1cs, proving and verifying keys) via the `Sync` opcode, which is introduced by this PR.

## Related Issue

https://community.provide.services/t/zkp-in-baseline-protocol/65

## Motivation and Context

Scalable ZKPs for production.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #415, opened 2021-11-08
Add reference implementations folder

skosito:
# Description

Please check relevant BLIP https://github.com/eea-oasis/baseline-blips/issues/5
I renamed examples to referenced-implementations and then run bri-1 and bri-2 readme commands and checked if i get same results. I did not do any testing yet for remaining (deprecated radish34 and lib folder in bri-1) since they seem more involved for verification, but I will try verify as well - if you have some advice here, please let me know.

Would be awesome if reviewers can also do quick bri-1/bri-2 checks. I  was getting socket hangout with bri-1, but same behavior was there even before changes.

Then I cleaned up mentions of examples folder (mostly in urls) across codebase and replaced it with reference-implementations. Should we revisit this and also leave examples folder in some .npmignore files since we will probably need it once we actually add examples folder again?

There is one more place that is referencing examples folder, in couple of json files in bri-2, but it seems like an local output of some tool (truffle?), in this format (this is Shield.json file in bri-2):
"sourcePath": /Users/samuelstokes/repos/Web3Studio/baseline/examples/bri-2/contracts/contracts/privacy/Shield.sol.
Do we need these types of files, since they seem connected to local build anyways?

## Related Issue

None

## Motivation and Context

Described in above mentioned BLIP. Mostly to avoid confusion, especially in future when we have examples and integrations folders.

## How Has This Been Tested

I ran bri-1 and bri-2 before changes and after. These folders seems isolated from rest of the app, and they should be isolated.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ x] My code follows the code style of this project.
- [x ] My change requires a change to the documentation.
- [x ] I have updated the documentation accordingly.
- [x ] All new and existing tests passed.


kthomas:
Would be good to consolidate this simple rename of `examples` -> `reference-implementations` in a single commit.

Also, we need to merge https://github.com/eea-oasis/baseline/pull/299 before applying this rename.

kthomas:
Also, under `bri-1`, I would do as you suggest-- that is, move everything under base-example/ up under the root directory... and keep `lib/`.

kthomas (dismissed):
See comments...

skosito:
> See comments...

Please check again, interactive rebasing this to squash commits was really tricky - based on a diff it seems none of the files was deleted. 
Could you run bri-1 again to check, for me it was hanging even before this, so it is impossible for me to verify on my machine.

GoldenBit0:
@kthomas @skosito What is the status of this PR?

ognjenkurtic (approved):

humbitious (approved):
loo

humbitious (approved):

skarred14 (approved):

kthomas:
On second thought, we may want to reconsider this change, revert it and have a short discussion about how to mitigate the problem that this change breaks _a lot_ of links across the web...

It would be good for those links to keep working... for example, I recall many press articles that linked to examples/bri-1 as well as examples/bri-2...

Users should not get a 404 to those links.

@humbitious ^ 👀

skosito:
> On second thought, we may want to reconsider this change, revert it and have a short discussion about how to mitigate the problem that this change breaks _a lot_ of links across the web...
> 
> It would be good for those links to keep working... for example, I recall many press articles that linked to examples/bri-1 as well as examples/bri-2...
> 
> Users should not get a 404 to those links.
> 
> @humbitious ^ 👀

@kthomas I am not sure why we need to revert PR? We can try to find solution without reverting it imo - do you have some suggestion? How much links there are around the web? How much of those we don't have control over to simply go and edit? I did quick search, and only thing I found is not up to date even without this change: https://docs.baseline-protocol.org/bri/bri-1/bri-1-base-example, here it still says you need to checkout `bri-1-privacy` branch, which is quite outdated as well...

humbitious:
As discussed in core dev channel, we can’t be held up from refactoring because of external hard links, and there was plenty of time to discuss alternatives before we pushed. However, Provide did do a ‘mea culpa’ on not reviewing well enough during the review period.  Willing to consider changing (THIS WEEK) to /examples where BRIs are (for now) still in examples, but where we put other examples in the /examples folder like so:  /examples/battleship.   @kthomas this will mean that all the various examples will be at the same folder level as BRIs. This will position any BRI simply as another example, and it will make them even harder to find than they already are, but…if the damage has been done, and there are too many external links to not hurt one of our contributing products in the short term, then this would be acceptable, at least to me, and it would allow us to unblock the examples, which we will simply put into the /examples folder directly as their own folders.  Happy to make a motion to consider this and execute…THIS WEEK.

kthomas:
> As discussed in core dev channel, we can’t be held up from refactoring because of external hard links, and there was plenty of time to discuss alternatives before we pushed. However, Provide did do a ‘mea culpa’ on not reviewing well enough during the review period. Willing to consider changing (THIS WEEK) to /examples where BRIs are (for now) still in examples, but where we put other examples in the /examples folder like so: /examples/battleship. @kthomas this will mean that all the various examples will be at the same folder level as BRIs. This will position any BRI simply as another example, and it will make them even harder to find than they already are, but…if the damage has been done, and there are too many external links to not hurt one of our contributing products in the short term, then this would be acceptable, at least to me, and it would allow us to unblock the examples, which we will simply put into the /examples folder directly as their own folders. Happy to make a motion to consider this and execute…THIS WEEK.

I reject this with prejudice.

Pleaser include @skarred14. To be honest I actually spent time reviewing this... I commented on it several times. I didn't have an allergic reaction at all to the changes until I realized after it had been merged. What if I had raised these valid concerns before it had been merged? In that case, is it Provide exerting its influence on this repo? Or is it someone else?

humbitious:
> > As discussed in core dev channel, we can’t be held up from refactoring because of external hard links, and there was plenty of time to discuss alternatives before we pushed. However, Provide did do a ‘mea culpa’ on not reviewing well enough during the review period. Willing to consider changing (THIS WEEK) to /examples where BRIs are (for now) still in examples, but where we put other examples in the /examples folder like so: /examples/battleship. @kthomas this will mean that all the various examples will be at the same folder level as BRIs. This will position any BRI simply as another example, and it will make them even harder to find than they already are, but…if the damage has been done, and there are too many external links to not hurt one of our contributing products in the short term, then this would be acceptable, at least to me, and it would allow us to unblock the examples, which we will simply put into the /examples folder directly as their own folders. Happy to make a motion to consider this and execute…THIS WEEK.
> 
> I reject this with prejudice.
> 
> Pleaser include @skarred14. To be honest I actually spent time reviewing this... I commented on it several times. I didn't have an allergic reaction at all to the changes until I realized after it had been merged. What if I had raised these valid concerns before it had been merged? In that case, is it Provide exerting its influence on this repo? Or is it someone else?

Unclear what you are rejecting here. The point about the mea culpa (which I intended as a commendation, not a sleight) or the attempt at a compromise where we keep BRIs in the /examples folder?

humbitious:
> > > As discussed in core dev channel, we can’t be held up from refactoring because of external hard links, and there was plenty of time to discuss alternatives before we pushed. However, Provide did do a ‘mea culpa’ on not reviewing well enough during the review period. Willing to consider changing (THIS WEEK) to /examples where BRIs are (for now) still in examples, but where we put other examples in the /examples folder like so: /examples/battleship. @kthomas this will mean that all the various examples will be at the same folder level as BRIs. This will position any BRI simply as another example, and it will make them even harder to find than they already are, but…if the damage has been done, and there are too many external links to not hurt one of our contributing products in the short term, then this would be acceptable, at least to me, and it would allow us to unblock the examples, which we will simply put into the /examples folder directly as their own folders. Happy to make a motion to consider this and execute…THIS WEEK.
> > 
> > 
> > I reject this with prejudice.
> > Pleaser include @skarred14. To be honest I actually spent time reviewing this... I commented on it several times. I didn't have an allergic reaction at all to the changes until I realized after it had been merged. What if I had raised these valid concerns before it had been merged? In that case, is it Provide exerting its influence on this repo? Or is it someone else?
> 
> Unclear what you are rejecting here. The point about the mea culpa (which I intended as a commendation, not a sleight) or the attempt at a compromise where we keep BRIs in the /examples folder?

Also, no — Provide making the case before merge would not be exerting undue influence. On the contrary. We may very well have said, “Oh…good points…let’s find another way that achieves the refactoring goal with less disruption.”  There is nothing wrong with realizing a problem only after the fact. Sometimes that’s how it goes. But then we discuss. You don’t revert unilaterally. We’re all pretty dialed in here, so if this was urgent, we could have called a meeting fast and hammered it out.

PR #469, opened 2022-03-25
Feature/blip 6 standard requierments updates

Ybittan:
# Description

This feature branch implements R217 into the BLIP6 BPI Monolith Test Suite

## Related Issue
BLIP6 

## Motivation and Context

This change implements a baseline standard requirement, maintaining its compliance. 

## How Has This Been Tested

The 13 tests within the test suite were executed several times, passing each time consistently. 
In addition, debugging with several breakpoints to validate data flowing through to ensure accuracy.

## Types of changes

## Checklist


Manik-Jain (dismissed):

examples/bpi-monolith-test-suite/bpi/components/vsm/vsm.ts
```diff
@@ -22,7 +22,7 @@ export class VirtualStateMachine implements IVirtualStateMachineComponent {
         const workgroup = this.workgroupComponent.getWorkgroupById(transaction.workgroupId);
         const workstep = workgroup.getWorkstepById(transaction.workstepId);
 
-        // If succesfull, this changes the agreement state
+        // If succesful, this changes the agreement state
```
skosito, line 25: typo: successful
nitpick: some empty space added at the end, when we add some linter, this might be caught as error

examples/bpi-monolith-test-suite/bpi/components/workgroup/workgroup.ts
```diff
@@ -6,16 +6,24 @@ export class Workgroup {
      name: string;
      participants: BpiSubject[] = [];
      worksteps: Workstep[] = [];
+     workstepids: Set<string> = new Set("");
```
skosito, line 9: let's maybe use camel case here `workstepIds`
also, please add a comment why workstepIds is needed in addition to worksteps, i think it might be useful

examples/bpi-monolith-test-suite/bpi/components/workgroup/workgroup.ts
```diff
@@ -6,16 +6,24 @@ export class Workgroup {
...
 
     constructor(name: string, id: string, owner: BpiSubject, worksteps: Workstep[],) {
-        this.name = name;
         this.id = id;
+        this.name = name;
         this.worksteps = worksteps;
```
skosito, line 14: since we are initializing `worksteps` here, should we also initialize `workstepIds` using `worksteps`? maybe simplest way for this would be
foreach workstep call addWorkstep?

examples/bpi-monolith-test-suite/tests/tests.ts
```diff
@@ -12,7 +12,7 @@ import { MESSAGE_TYPE_INFO, MESSAGE_TYPE_STORE } from '../bpi/components/messagi
 describe('BPI, Workgroup and Worflow setup', () => {
     it('Given beggining of time, Alice initiates a new BPI with her organization as owner and an inital agreement state object, BPI created with inherent owner and a global agreement', () => {
         const bpiInstance = new BPI("AL1", "AliceOrganisation", ["555333"], new IdentityComponent(), new MockMessagingComponent(), new WorkgroupComponent());
-
+
```
skosito, line 15: same "nitpick instead of linter" comment :)

examples/bpi-monolith-test-suite/bpi/components/workgroup/workgroup.ts
```diff
@@ -6,16 +6,24 @@ export class Workgroup {
...
         this.participants.push(owner);
     }
 
      addWorkstep(workstep: Workstep) {
-          this.worksteps.push(workstep);
+
+          //push workstep if ID is not yet in use
+          if(!(this.workstepids.has(workstep.id))) {
```
skosito, line 21: i dont think we need parenthesis after `!`

also, a bit unclear here, why multiple worksteps can have same id? does this mean that addWorkstep can be called multiple times for same workstep? how much worksteps there can be, if it's usually small number maybe adding set for this to optimize search might not be worth and you can just use getWorkstepById? maybe we can rename function as `addWorkstepIfNotExists`, and simplify it like:

if (this.workstepIds.has(workstep.id)) {
return;
}

this.worksteps.push(workstep);
this.workstepids.add(workstep.id);

examples/bpi-monolith-test-suite/bpi/components/workgroup/workgroup.ts
```diff
@@ -6,16 +6,24 @@ export class Workgroup {
...
+               this.worksteps.push(workstep);
+          }
+
+          //add ID to running set
+          this.workstepids.add(workstep.id);
      }
 
      getWorkstepById(workstepId: string) {
```
skosito, line 29: i know it is not related to your changes, but there can be a bug in this function if workstep with this id doesn't exist, same for  getParticipantsById, so we can quickly fix this now i think, something like this should work:
`return worksteps.length >0 && worksteps[0]`

skosito (commented):
Left some comments, mostly small things, please let me know if they make sense

examples/bpi-monolith-test-suite/bpi/components/workgroup/workgroup.ts
Ybittan: The reason for these changes is to prevent someone from reusing a previously used ID in a workgroup. Per R217 this would be in violation of the standard, so my changes here is to make it impossible for an ID to be reused, by tracking the IDs in current use in the set. I considered using getWorkstepById, but I worried about search time complexity. In this small test suite it is fast either way, and not much of a concern, but I worried that the scalability is lost, and wanted a solution that can be used for any implementation of any size. What are your thoughts on this?

examples/bpi-monolith-test-suite/bpi/components/workgroup/workgroup.ts
```diff
@@ -6,21 +6,30 @@ export class Workgroup {
...
      addWorkstep(workstep: Workstep) {
-          this.worksteps.push(workstep);
+
+          //push workstep if ID is not yet in use
+          if(!this.workstepIds.has(workstep.id)) {
+               this.worksteps.push(workstep);
+          }
+
+          //add ID to running set
+          this.workstepIds.add(workstep.id);
```
skosito, line 27: @Ybittan thanks for answer, I get that Set is used to make sure id is only used once. But in my opinion using Set here is overkill,  you can just ask if workstep exist, it will always be small enough set in this implementation - just my opinion, feel free to ignore :) 

But what happens if addWorkstep is called with existing id, this function will just ignore that? In that case, caller of this function would not be aware that his workstep was not added because he is using existing id. Some error should be returned.

I think it would be better to make sure ID is always unique when creating workstep, by using uuid for example, instead of checking it like this.

examples/bpi-monolith-test-suite/bpi/components/workgroup/workgroup.ts
skosito: @Ybittan reminder about this comment :)

skosito (approved):

fosgate29 (approved):

PR #507, opened 2022-08-25
Feature/489 bri 3 accounts model

skosito:
# Description

According to document referenced in linked issue https://github.com/eea-oasis/baseline/issues/489. This covers first task, adding simple model, more implementation will come through CRUD task following up soon.

## Related Issue

https://github.com/eea-oasis/baseline/issues/489

## Motivation and Context

Part of BRI-3 (simple reference implementation) effort.

## How Has This Been Tested

Will be tested with CRUD task, for now it is just simple model.

## Types of changes

## Checklist


examples/bri-3/components/identity/bpiAccount/bpiAccount.ts
```diff
@@ -0,0 +1,12 @@
+import { IBpiAccount } from "./bpiAccount.interface";
+
+export class BpiAccount implements IBpiAccount {
+    id: string; // TODO: Add uuid after #491
+    nonce: number;
+    ownerBpiSubjectId: string;
+
+    constructor(nonce: number, ownerBpiSubjectId: string) {
+        this.nonce = nonce;
```
skosito, line 9: we will probably add better way for nonce initialization, maybe some function that increments it, but for now maybe its enough to have it in constructor?

fosgate29: It is enough to have it in constructor I believe

fosgate29 (dismissed):
lgtm

examples/bri-3/components/identity/bpiAccount/bpiAccount.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface IBpiAccount {
+    id: string; // TODO: Add uuid after #491
+    nonce: number;
+    ownerBpiSubjectId: string;
```
Therecanbeonlyone1969, line 4: that should be an array in order to allow for flexibility and extensibility. Plus typically, a BPI account will have two owners since it connects to a joint state.

examples/bri-3/components/identity/bpiAccount/bpiAccount.ts
```diff
@@ -0,0 +1,12 @@
+import { IBpiAccount } from "./bpiAccount.interface";
+
+export class BpiAccount implements IBpiAccount {
+    id: string; // TODO: Add uuid after #491
+    nonce: number;
+    ownerBpiSubjectId: string;
```
Therecanbeonlyone1969, line 6: see previous comment

examples/bri-3/components/identity/bpiAccount/bpiAccount.ts
```diff
@@ -0,0 +1,12 @@
+import { IBpiAccount } from "./bpiAccount.interface";
+
+export class BpiAccount implements IBpiAccount {
+    id: string; // TODO: Add uuid after #491
+    nonce: number;
```
Therecanbeonlyone1969, line 5: since there is more than one way to generate a deterministic nonce a string might be more appropriate

Therecanbeonlyone1969 (changes requested):
See comments

examples/bri-3/components/identity/bpiAccount/bpiAccount.interface.ts
ognjenkurtic: Could you please elaborate on the second sentence. Why should it have two owners typically? Isn't it more common that org A would work with one BpiAccount and org B with another?

ognjenkurtic: I agree that it should be an array, multiple owners makes sense in multisig scenarios.

ognjenkurtic (commented):
I am ok with the PR as soon as @Therecanbeonlyone1969 regarding multiple owners is addresed

examples/bri-3/components/identity/bpiAccount/bpiAccount.interface.ts
Kasshern: Perhaps we should we should discuss and conclude a scope of functionality for alignment of team and maintainers in future PRs. Originally the team agreed on the "simplest implementation" to cover our intended use case. We seem to ever so slowly be drifting towards accommodating common use case needs that fall outside the "MUST" requirements of the standard; a direction in which has my support.

Kasshern (commented):
I concur with previous comments and approve pending suggested changes of @Therecanbeonlyone1969

examples/bri-3/components/identity/bpiAccount/bpiAccount.interface.ts
Therecanbeonlyone1969: If we want the SRI to be useful for multi party coordination WITHOUT being clunky aka you need to double or triple the BPI accounts. And will impact adoption. 

I agree we should keep it as simple as possible while ensure good usability.

examples/bri-3/components/identity/bpiAccount/bpiAccount.ts
Manik-Jain: Considering the upper limit of TS number type, I think we shall implement it as bignumber right since the beginning.

examples/bri-3/components/identity/bpiAccount/bpiAccount.ts
Manik-Jain: @skosito ideally the constructor caller function must be having the nonce increment logic. So I think it is fine to have the constructor initializer

Manik-Jain:
> I concur with previous comments and approve pending suggested changes of @Therecanbeonlyone1969

+1

ognjenkurtic (approved):

Kasshern (approved):

Therecanbeonlyone1969 (approved):
LGTM

Issue #491, opened 2022-08-02
Decide on and introduce a server framework for SRI


Therecanbeonlyone1969:
## Overview

Core Devs need to decide on the server framework used for the SRI nad implement it as part of the node.js project.

### Questions

- use express vs koa vs axios?

### Assumptions

- using node.js

### Acceptance

- Framework decided
- Node.js project setup
- Server framework introduced with a dummy endpoint

## Tasks

- [x] Decide on Framework components
- [x] Setup node.js project
- [x] IMplement dummy endpoint with the selected server framework

skosito:
My suggestion is we use express.js, it is very simple and common choice.

Therecanbeonlyone1969:
question: given that the SRI will probably be most commonly used to not only get started but also as the basis for company-specific implementation, are there any downsides to using express vs axios vs koa for example?

Manik-Jain:
Express can act as a quick starter as @skosito mentioned, and considering what @Therecanbeonlyone1969 has mentioned I think it would be even better to have some out of the box implementation frameworks like NestJS that build upon the basic express and TS, and provide improved package control.

https://nestjs.com/

ognjenkurtic:
@Manik-Jain i agree, we do not need to reinvent the wheel here. I especially like https://docs.nestjs.com/recipes/cqrs. Proven scalable pattern for a growing codebase. I would be happy to implement an example and share with the team.

@Therecanbeonlyone1969 i am not an expert on node. Express and Koa are server frameworks while Axios is a HTTP client. Koa is marketed as the spiritual leader of Express. but i am not sure if our choice would make any impact on the adoption. I am much more in favor of Manik's proposal - it means we can bootstrap quickly and focus on the architecture and features, instead of on the low level stuff. NestJS is a commonly used framework that should be understandable to a large number of devs.

ognjenkurtic:
> @Manik-Jain i agree, we do not need to reinvent the wheel here. I especially like https://docs.nestjs.com/recipes/cqrs. Proven scalable pattern for a growing codebase. I would be happy to implement an example and share with the team.
> 
[...]

Issue #492, opened 2022-08-02
Introduce a linter and formatting style for the SRI project


ognjenkurtic:
## Overview

We need this to make sure formatting is uniform across the project.

### Reference

- [Follows from this issue](https://github.com/eea-oasis/baseline/issues/491)

### Questions

- Linter?
- Style?

### Acceptance

- Linter introduced
- Formatting style agreed and introduces

## Tasks

- [ ] Linter setup and running locally

Kasshern:
I vote for typescript-eslint and a modified version of prettier.

Manik-Jain:
> I vote for typescript-eslint and a modified version of prettier.

+1 for typescript-eslint. I think default prettier does pretty good job, is there a reason for the proposed modification?

skosito:
Its probably best to go with default options and no (or very minial) config in both linter and prettier for start

Kasshern:
You make a very good point @skosito. You have convinced me, albeit reluctantly, to adopt default options for both the linter and prettier.

PR #508, opened 2022-08-27
Feature/491 nestjs cqrs poc

ognjenkurtic:
# Description

Setup of the NestJs project with CQRS architecture. Implements an example of Create operation for a Bpi Subject with the CQRS flow. Implements an example test suite of the Subjects controller.

Contains a lot of TODOs for future work, and comments that explain the architecture and that will be deleted as everyone gets onboard with the approach.

## Related Issue

#491 

## Motivation and Context

It introduces a  server framework and architecture concepts that will be the basis for all future work.

## How Has This Been Tested

Ran locally. Ran unit tests. 

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist

- [ x] My code follows the code style of this project.
- [x ] I have added tests to cover my changes.
- [ x] All new and existing tests passed.


examples/bri-3/src/bri/identity/bpiSubjects/models/bpiSubject.interface.ts
```diff
@@ -1,5 +1,6 @@
 import { BpiSubjectType } from "./bpiSubjectType.enum";
 
+// TODO: Do we actually need interfaces as we are already working with classes?
```
Therecanbeonlyone1969, line 3: @ognjenkurtic If you want to create and pass a type-checked class object, you should use ts classes. If you need to work without creating an object, an interface is best.

Use an interface for the following:
- Validating specific structure of properties
- Objects as parameters
- Objects returned from functions

So we need to determine if we want to use classes but not instantiate them, then we combine interfaces with classes.

Get's off his soapbox

Therecanbeonlyone1969 (approved):
great start... we need to document the usage of interfaces vs classes. I think this should be driven by whether the class is stateless or stateful ... for stateful you probably want to instantiate the object and load the data into it, vs invoking a service that just processes data that can be invoked at run time.

Or am I missing something?

@ognjenkurtic @skosito @Kasshern @Manik-Jain

Manik-Jain (approved):

examples/bri-3/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    'plugin:prettier/recommended',
+  ],
+  root: true,
+  env: {
+    node: true,
+    jest: true,
+  },
+  ignorePatterns: ['.eslintrc.js'],
+  rules: {
+    '@typescript-eslint/interface-name-prefix': 'off',
```
skosito, line 20: is turning these off by default or some custom rules?

examples/bri-3/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Stay in touch
+
+- Author - [Kamil Myśliwiec](https://kamilmysliwiec.com)
+- Website - [https://nestjs.com](https://nestjs.com/)
+- Twitter - [@nestframework](https://twitter.com/nestframework)
+
+## License
+
+Nest is [MIT licensed](LICENSE).
```
skosito, line 73: even though we don't need readme that describes project fully, i think removing this nestjs boilerplate from readme would be great

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { BadRequestException } from '@nestjs/common';
```
skosito, line 1: small thing but just reminder that we should maybe discuss if we should unit test agents or test them through controllers like this only, or both?

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.ts
```diff
@@ -0,0 +1,18 @@
...
+import { CommandBus } from '@nestjs/cqrs';
+import { CreateBpiSubjectCommand } from '../capabilities/createBpiSubject/createBpiSubject.command';
+import { BpiSubject } from '../models/bpiSubject';
+import { CreateBpiSubjectDto } from './dtos/request/createBpiSubject.dto';
+
+@Controller("subjects")
+export class SubjectController {
+  constructor(private commandBus: CommandBus) {}
+
+  // TODO: DTO validation
```
skosito, line 11: what we will validate on this level and what on agent level?

examples/bri-3/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "eslint-plugin-prettier": "^4.0.0",
+    "jest": "28.1.2",
+    "prettier": "^2.3.2",
+    "source-map-support": "^0.5.20",
+    "supertest": "^6.1.3",
+    "ts-jest": "28.0.5",
+    "ts-loader": "^9.2.3",
+    "ts-node": "^10.0.0",
+    "tsconfig-paths": "4.0.0",
+    "typescript": "^4.3.5"
```
skosito, line 53: since the project is just starting i think we should go with latest typescript from the start

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/createBpiSubject/createBpiSubjectCommand.handler.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    this.agent.throwIfCreateBpiSubjectInputInvalid(name, desc, publicKey);
+
+    const newBpiSubject = this.agent.createNewExternalBpiSubject(name, desc, publicKey);
+    
+    // TODO: Generic map of domain model to entity model
+    // this.orm.store(newBpiSubject);
+
+    // TODO: Response DTO
+    return true;
```
skosito, line 20: is this return needed for commands?

examples/bri-3/src/bri/identity/bpiSubjects/models/bpiSubject.interface.ts
skosito: to me now these interfaces seem redundant, could change i guess

examples/bri-3/src/bri/identity/bpiSubjects/models/bpiSubject.ts
```diff
@@ -2,6 +2,7 @@ import { IBpiSubject } from "./bpiSubject.interface";
 import { BpiSubjectType } from "./bpiSubjectType.enum";
 
 export class BpiSubject implements IBpiSubject {
+    // Fields of the domain model are always changed through it's methods
```
skosito, line 5: maybe let's already make them private?

Kasshern (approved):
@ognjenkurtic This PR seems to include all the necessary changes for introducing configuration of linter and prettier format files. I advise linking corresponding issue #492 here, placing said issue in your name, and clearing it upon merge of this PR.

examples/bri-3/src/bri/workgroup/workgroups/models/workgroup.ts
```diff
@@ -1,16 +1,16 @@
...
+import { Security } from '../../../policy/models/security';
+import { Privacy } from '../../../policy/models/privacy';
 
 export class Workgroup implements IWorkgroup {
     id: string; // TODO: Add uuid after #491
     name: string;
     administrator: BpiSubject[] = []; 
-    securityPolicy: Security; //TODO Implement securityPolicy #485
-    privacyPolicy: Privacy; //TODO Implement privacyPolicy #485
+    securityPolicy: Security[]; //TODO Implement securityPolicy #485
```
skosito, line 12: some of these fields are arrays but name is singular, for example administrator above is array, policy here is array etc, just to double check if we should use plural for names? why are these modified to arrays in this PR?

examples/bri-3/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "declaration": true,
+    "removeComments": true,
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "allowSyntheticDefaultImports": true,
+    "target": "es2017",
```
skosito, line 9: i think we should go with highest possible target here, even es2022, is there a reason to go with es2017?

skosito (approved):
lgtm, some minor comments, feel free to ignore those that don't make sense

examples/bri-3/.eslintrc.js
ognjenkurtic: default. we can agree on the rules as part of #492

examples/bri-3/src/bri/identity/bpiSubjects/models/bpiSubject.interface.ts
ognjenkurtic: Interfaces are redundant for me as well. We want to instantiate objects and these objects will encapsulate all the logic for their state changing. Let's discuss in a separate issue, not a blocker for now.

examples/bri-3/.eslintrc.js
ognjenkurtic: > great start... we need to document the usage of interfaces vs classes. I think this should be driven by whether the class is stateless or stateful ... for stateful you probably want to instantiate the object and load the data into it, vs invoking a service that just processes data that can be invoked at run time.
> 
> Or am I missing something?
> 
> @ognjenkurtic @skosito @Kasshern @Manik-Jain

Thx. See my answer above re: interfaces vs classses. Classes seems like the best option for domain objects. We ll continue the discussion in parallel as we move forward.

ognjenkurtic:
> @ognjenkurtic This PR seems to include all the necessary changes for introducing configuration of linter and prettier format files. I advise linking corresponding issue #492 here, placing said issue in your name, and clearing it upon merge of this PR.

They are just introduced as part of this PR. I would leave #492 open so that someone get the team to agree on the ruleset amd then configure them properly.

examples/bri-3/README.md
ognjenkurtic: Yes, will do as a next step, didnt have time to do it as part of this one.

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.spec.ts
ognjenkurtic: Open for recommendations. For me this combinations seems reasonable:
1) Unit test controllers with mocked storage (covers all the business logic)
2) E2E test endpoints (tests the full system without mocking)

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.ts
ognjenkurtic: On this level we validate the input - request DTO. For example foo must be string and bar must not be emtpy. 
On the agent level we validate business logic. For example, entity X cannot be edited if in status Y.

examples/bri-3/package.json
ognjenkurtic: Will bump the version as a first next step

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/createBpiSubject/createBpiSubjectCommand.handler.ts
ognjenkurtic: I just added for the case of the example. I do not think it is obligatory. I believe nestjs would by default return 200 in case no exceptions thrown (this needs to be validated in nestjs docs)

examples/bri-3/src/bri/identity/bpiSubjects/models/bpiSubject.ts
ognjenkurtic: we cant as long as we have interfaces as they do not allow for access modifiers.

examples/bri-3/tsconfig.json
ognjenkurtic: This was default due to my version of node when setting up nestjs. Will bump as first next step.

examples/bri-3/src/bri/workgroup/workgroups/models/workgroup.ts
ognjenkurtic: I think we are still far from implementing these so no need to touch them atm. They will be handled as part of other stories and are here only as placeholders for now.

Issue #484, opened 2022-07-21
SRI Zero-Knowledge Component: Off-chain Merkle Tree


biscuitdey:
## Overview

Implement an off-chain [Merkle Tree class](https://docs.google.com/document/d/1FTcW1kPa-75AqaYLdlPeL2TXNR-NJy_nN1ZUvgk-m18/edit?usp=sharing) to convert state documents into a Merkle Tree and return the root.

### Reference

- [Documentation](https://docs.google.com/document/d/1FTcW1kPa-75AqaYLdlPeL2TXNR-NJy_nN1ZUvgk-m18/edit?usp=sharing)
- This implementation uses the [precise-proof library](https://github.com/energywebfoundation/precise-proofs) to allow complex documents to be converted into a Merkle Tree.

### Assumptions

- Documents to be converted into Merkle Tree store data in a key-value format

### Acceptance

- Return a Merkle Tree Root

## Tasks

- [x] Calculate Merkle Tree taking document key-value pair as leaves (input)
- [x] Return the root of the Merkle Tree

PR #509, opened 2022-08-31
Feature/484-bpi-3-zkp-component

biscuitdey:
# Description

According to document referenced in linked issue #484. This covers the task of creating a Merkle tree and getting its root.

## Related Issue

#484

## Motivation and Context

Part of BRI-3 (simple reference implementation) effort.

## How Has This Been Tested

1. Tested for being able to create Merkle tree.
2. Tested for being able to return the Merkle tree root.
3. No collision in the Merkle tree root created.

## Types of changes

## Checklist


examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -0,0 +1,74 @@
...
+  tree: Tree;
+
+  constructor(document: any){
+    const merkleTree: Tree = this.createTree(document);
+
+    this.tree.leaves = merkleTree.leaves;
+    this.tree.root = merkleTree.root; 
+  }
+
+  private createTree(document: any): Tree {
```
Manik-Jain, line 15: please add early exit if document is empty. Also, is there a specific reason to use "any" type for document?

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -0,0 +1,74 @@
...
+    try {
+      var tree: Tree;
```
Manik-Jain, line 17: please utilize `let`

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -0,0 +1,74 @@
...
+
+      const leaves: PreciseProofs.Leaf[] = this.createLeaves(document);
```
Manik-Jain, line 19: what are the chances that leaves can be undefined here?

Manik-Jain (commented):
Added minor comments, looks good overall

examples/bri-3/src/bri/zkp/components/merkleTree/tree.d.ts
```diff
@@ -0,0 +1,6 @@
+import { PreciseProofs } from "ew-precise-proofs-js";
+
+export type Tree = {
```
skosito, line 3: why using type here instead of interface?

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
skosito: i think you can just init class variable tree inside createTree method (maybe call it initTree), so you don't need separate tree variable here, but just a preference
if you leave it like this, you dont need to declare tree here you can just return it and declare there

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -0,0 +1,74 @@
+import { Tree } from "./tree"
+import { IMerkleTree} from "./merkleTree.interface";
+import { PreciseProofs } from "ew-precise-proofs-js";
```
skosito, line 3: could you please explain usage of this package and why it will be useful for us, currently it is just for merkle tree creation, but i assume that will change?

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -0,0 +1,74 @@
...
+      );
+
+      const root = PreciseProofs.getRootHash(merkleTree);
+      
+      tree.leaves = leaves;
+      tree.root = root;
+
+      return tree;
+    } catch (e) {
+      console.log("Tree not created");
```
skosito, line 32: nitpick but lets use one log console.log(`Tree not created: ${e}`)

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -0,0 +1,74 @@
+import { Tree } from "./tree"
+import { IMerkleTree} from "./merkleTree.interface";
+import { PreciseProofs } from "ew-precise-proofs-js";
+
+export class MerkleTree implements IMerkleTree {
+  tree: Tree;
+
+  constructor(document: any){
+    const merkleTree: Tree = this.createTree(document);
```
skosito, line 9: since createTree is catching error and not returning anything this method will return undefined if error occurs during tree creation, which will throw error on lines 11 and 12, maybe modify createTree method to return empty tree in case of an error, or add some if (merkleTree) guard here

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -0,0 +1,74 @@
...
+      return tree;
+    } catch (e) {
+      console.log("Tree not created");
+      console.log(e);
+    }
+  }
+
+  private createLeaves(document: any): PreciseProofs.Leaf[] {
+    try {
+      const leaves = PreciseProofs.createLeafs(document);
```
skosito, line 39: return PreciseProofs.createLeafs(document)
in case of an error, same thing as createTree, returning undefined for leaves here might lead to some bugs, we will need either guards everywhere we use leaves or returning default here
btw, i wonder why this library is using wrong plural for leaf? :)

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -0,0 +1,74 @@
...
+      console.log("Leaves not created");
+      console.log(e);
+    }
+  }
+
+  getTree(): Tree {
+    try {
+      return this.tree;
+    } catch (e) {
+      console.log("Unable to get tree");
```
skosito, line 51: i dont think return this.tree can throw exception, please remove try and catch
also default visibility in ts for class members is public, so `tree` is accessible even without this getter, maybe lets make tree private

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -0,0 +1,74 @@
...
+      console.log(e);
+    }
+  }
+
+  getLeaves(): PreciseProofs.Leaf[]{
+    try {
+      return this.tree.leaves;
```
skosito, line 58: same comment for redundant try and catch, also redundant getLeaves() and getRoot() methods since getTree().leaves and getTree().root are doing the trick already

examples/bri-3/test/zkp/merkleTree.test.ts
```diff
@@ -0,0 +1,42 @@
+import { MerkleTree } from "../../src/bri/zkp/components/merkleTree/merkleTree";
+
+describe("Create a Merkle Tree and return a unique root", () => {
```
skosito, line 3: can we add tests for leaves too? i can help with adding those tests so i can also learn about library a bit if you don't mind @biscuitdey ?

skosito (changes requested):
looks good, but some small tweaks and comments to be addressed before merge

package.json
```diff
@@ -1 +1,9 @@
-{}
\ No newline at end of file
+{
```
ognjenkurtic, line 1: This package.json is not in the correct location. You should update existing package.json in examples/bri-3/

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
ognjenkurtic: in case leaves is undefined, leaves.map exception will be caught by the catch block bellow. In that case , this method should return something as Skos said (i.e undefined) and the caller then must check the return value before trying to assign values from it.

examples/bri-3/test/zkp/merkleTree.test.ts
ognjenkurtic: +1

ognjenkurtic (changes requested):

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.interface.ts
```diff
@@ -0,0 +1,6 @@
+import { Tree } from "./tree"
+export interface IMerkleTree {
+  tree: Tree;
+  getRoot: (document: any) => string;
+}
```
Therecanbeonlyone1969, line 5: will need more functions here. please add a TODO here @biscuitdey

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
Therecanbeonlyone1969: @skosito we will be storing zkp witnesses in merkle trees. also we will use merkle trees to record BPI state associated with a BPI account. similar to how ethereum does it ... one could ask the question if we should use Patricia-Merkle trees vs regualr merkle trees but for now this is ok, IMHO.

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
Therecanbeonlyone1969: @biscuitdey i agree with @skosito ... please add. thank u

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
Therecanbeonlyone1969: @skosito you are actually bringing up a good point.

Could we please NOT USE console.log but rather a ts logger e.g. https://www.npmjs.com/package/typescript-logger ... it will make troubleshooting so much easier. Let's agree on an approach and method!! cc @ognjenkurtic @Kasshern @Ybittan @Manik-Jain @biscuitdey

examples/bri-3/src/bri/zkp/components/merkleTree/tree.d.ts
Therecanbeonlyone1969: same thought

examples/bri-3/test/zkp/merkleTree.test.ts
Therecanbeonlyone1969: +1

Therecanbeonlyone1969 (changes requested):
See comments

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
ognjenkurtic: Agree completely, console.log should be used only for local development and debugging purposes. Let's discuss the proper approach for logging in a separate issue, i ll open one.

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
biscuitdey: Currently, the Merkle tree library I am using supports documents in JSON format and is projected to support other formats in the future. Hence, I have kept the document type as "any".

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { MerkleTree } from './merkleTree';
+
+describe('merkleTree', () => {
+  it('Should create merkle tree out of empty document', () => {
+    const document = {};
+
+    const merkleTree = new MerkleTree(document);
```
skosito, line 7: this is possible with precise-proofs so i just added test to confirm behavior

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -0,0 +1,39 @@
+import { ITree } from './tree.interface';
+import { PreciseProofs } from 'ew-precise-proofs-js';
+
+export class MerkleTree {
+  private tree: ITree = { leaves: [], root: undefined };
```
skosito, line 5: can MerkleTree class contain just leaves and root fields, why it needs tree object inside merkle tree?

skosito (commented):
hey @biscuitdey i did some cleanup because tests did not run and there were compilation issues and issues with logger library
logger is removed for now, i opened separate issue, in general libraries that affect whole repo, no matter how small they seem, should be handled through documented issue

please respond to my 2 new comments and then its approved from my side

Therecanbeonlyone1969 (commented):
Agree with @skosito ... modulo the tree object question, lgtm.

skosito (approved):

ognjenkurtic (approved):

Kasshern (approved):

Issue #488, opened 2022-07-26
SRI Subjects Component- Interfaces, Classes and API


ognjenkurtic:
## Overview

Create SRI Subject Component in the typescript language utilizing the Subjects and Accounts Component Requirements document as an outline of said component’s functionalities and structure.

### Reference

- [Reference](https://docs.google.com/document/d/1DCi0Ap9EK2cBxxEVzmOPkdkcl1hnv7-qoELGHODb1PM/edit)

Assumptions
- A portion of necessitated componential APIs, functionalities, and structures likely lay undocumented/undiscovered
- Cross-component integration is not considered within the scope of this issue

### Acceptance

 - Subject Component Interface created according to preparation document specification
 - Subject Component class inheriting interface created according to preparation document specification
 - Subject Component associated minor objects, model interfaces, and classes created as adhering to the preparation document specification
 - Subject Component associated utilities and functions created as adhering to guidelines inherently established through preparation document specification
 - Unit tests assessing Subject Component functionality, structure and instantiation created as covering at least 80% of viable code

## Tasks

- [x] Implement BPI Subject model  through interface and class that implements it
- [ ] Implement BPI Subject CRUD API (without talking to storage atm)
- [ ] Implement tests for above functionalities

PR #519, opened 2022-09-14
Feature/488 bpi subjects read update and delete

ognjenkurtic:
# Description

This PR introduces Update, Delete and Get ALl endpoints for BPI Subjects and closes completely #488

## Related Issue

#488 

## Motivation and Context

Introduces first full basic CRUD for one type of BRI entity 

## How Has This Been Tested

Added unit tests. Manual testing through submitted postman collection

## Types of changes

## Checklist


examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
```diff
@@ -1,22 +1,58 @@
...
+  public async fetchUpdateCandidateAndThrowIfUpdateValidationFails(id: string, name: string, desc: string, pk: string): Promise<BpiSubject> {
+
+    const bpiSubjectToUpdate = await this.repo.getBpiSubjectById(id);
+
+    if(!bpiSubjectToUpdate) {
+      throw new NotFoundException(`Bpi Subject with id: ${id} does not exist.`)
+    }
+
+    // Example additional checks
+    if (!name) {
```
skosito, line 35: couple nitpicks here:
- check if name is not defined before checking db seems cleaner
- empty lines on 27 and 49
- unclear comment on 34, not sure what example aditional check means
- i think desc and pk were also mandatory

examples/bri-3/src/bri/identity/bpiSubjects/api/dtos/request/updateBpiSubject.dto.ts
```diff
@@ -1,4 +1,5 @@
 export interface UpdateBpiSubjectDto {
+    id: string;
```
skosito, line 2: i dont think id should be part of update dto but rather part of api route

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.spec.ts
```diff
@@ -32,6 +39,84 @@ describe('SubjectController', () => {
...
+      // Act
+      const bpiSubjects = await sController.getAllBpiSubjects();
+
+      // Assert
+      expect(bpiSubjects.length).toEqual(0);
+    });
+
+    it('should return 1 bpi subject if 1 exists ', async () => {
+      // Arrange
+      const requestDto = { name: "name", desc: "desc", publicKey: "publicKey"} as CreateBpiSubjectDto;
```
skosito, line 81: can we mock db and return values for specific tests or we have to create objects like this?
also i think that test that returns 1 subject is redundant if we also have a test that returns 2 subjects below

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.spec.ts
```diff
@@ -55,31 +140,58 @@ describe('SubjectController', () => {
     });
   });
 
-  describe('getBpiSubjectById', () => {
+  describe('updateBpiSubject', () => {
```
skosito, line 143: do we need tests with empty name (description and pk)?

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.ts
```diff
@@ -27,13 +39,17 @@ export class SubjectController {
   }
 
   @Put()
```
skosito, line 41: id should be part of the route and not dto to be restful i believe, same as you did for delete

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/deleteBpiSubject/updateBpiSubjectCommand.handler.ts
```diff
@@ -0,0 +1,16 @@
+import { CommandHandler, ICommandHandler } from "@nestjs/cqrs";
+import { BpiSubjectAgent } from "../../agents/bpiSubjects.agent";
+import { BpiSubjectRepository } from "../../persistence/bpiSubjects.repository";
+import { DeleteBpiSubjectCommand } from "./deleteBpiSubject.command";
+
+@CommandHandler(DeleteBpiSubjectCommand)
```
skosito, line 6: file name here is wrong (update...) and execute method is doing delete

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/updateBpiSubject/updateBpiSubjectCommand.handler.ts
```diff
@@ -1,23 +1,19 @@
...
 
-    // const bpiSubjectToUpdate = this.agent.fetchBpiSubjectToUpdate(id);
-    // this.agent.updateBpiSubject(bpiSubjectToUpdate, name, desc, publicKey);
-
-    // TODO: Generic map of domain model to entity model
-    // this.orm.store(newBpiSubject);
-
-    // TODO: Response DTO
-    return true;
+    await this.repo.updateBpiSubject(bpiSubjectToUpdate);
```
skosito, line 15: just for discussion purposes it seems that there is no clear line if agent is doing db calls, or just validation etc
for example you could write this method as 

this.repo.getSubject()
this.agent.validateSubject()
this.repo.updateSubject()

or

this.agent.getSubjectAndValidate();
this.agentUpdate() which would also call repo and you can call method updateBpiSubjectAndThrowIfItFails ...

or couple other combinations like implementation in PR

dont have hard opinion on this, but curious on your thoughts about clear responsibilities

examples/bri-3/src/bri/identity/bpiSubjects/models/bpiSubject.ts
```diff
@@ -41,4 +40,16 @@ export class BpiSubject {
   public get publicKey(): string {
     return this._publicKey;
   }
+
+  public updateName(newName: string): void {
```
skosito, line 44: not hard opinion on this, but why not make name, description and publickkey public fields? it would eliminate 6 methods and will have same behavior

examples/bri-3/src/bri/identity/bpiSubjects/persistence/bpiSubjects.repository.ts
```diff
@@ -34,4 +46,29 @@ export class BpiSubjectRepository extends PrismaService {
...
+        return new BpiSubject(
+            newBpiSubjectModel.id, 
+            newBpiSubjectModel.name, 
+            newBpiSubjectModel.description, 
+            newBpiSubjectModel.type, 
+            newBpiSubjectModel.publicKey);
+    }
+
+    async deleteBpiSubject(bpiSubject: BpiSubject): Promise<void> {
+        await this.bpiSubject.delete({
```
skosito, line 70: this is not returned and function is marked to have Promise<void> return type, or i am missing something

examples/bri-3/test/bri.postman_collection.json
```diff
@@ -0,0 +1,158 @@
...
+					"port": "3000",
+					"path": [
+						"subjects"
+					]
+				}
+			},
+			"response": []
+		},
+		{
+			"name": "http://localhost:3000/subjects Copy",
```
skosito, line 37: name contains Copy at the end, below Copy 2, probably by mistake, also in get all no need for `all` in the name should be clear from rest route

skosito (approved):
left some comments, looks good, but please address comments before merge, i am approving to not block it

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
ognjenkurtic: Removed the check completely. It is just an example but it is confusing. These input fields will be checked by the input DTO validation, once we implement it. In the agent, we ll add additional validation of the business rules, once we get to them.

examples/bri-3/src/bri/identity/bpiSubjects/api/dtos/request/updateBpiSubject.dto.ts
ognjenkurtic: You are completely right. Fixed.

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.spec.ts
ognjenkurtic: Prisma does not support DB mocking to my knowledge. We could try to get the instance of the mock repo in the test and then fill it directly during Arrange. Will try that as part of some upcoming PR and will refactor if ok.

ognjenkurtic: Removed obsolete test

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.spec.ts
ognjenkurtic: Laters, once we introduce input DTO validation. Keep in mind that we are still in the early phase and most of the code is merged so that others can follow with their implementations.

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.ts
ognjenkurtic: Fixed

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/deleteBpiSubject/updateBpiSubjectCommand.handler.ts
ognjenkurtic: Fixed.

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/updateBpiSubject/updateBpiSubjectCommand.handler.ts
ognjenkurtic: Very good point. Renamed repo to agent. Explanation bellow:

We are early in the implementation so we do not have enough good examples in code but a good way to think about it is the following: 

Agent represents a person in an organization, or even better a role. 

Think about a hotel:

Each role is tasked with specific responsibilities. Lift boy will operate the elevator, receptionists will  welcome the guests and serve them at the reception...

In case that a person in a role is overloaded by too much work and\or too much responsibilities, organization would introduce additional persons and potentially additional roles, to help solve the problem. Think of it as horizontal and vertical scaling. Receptionist could have his\her role relieved by introducing another role (i.e. Cashier) that deals only with payments from the guests, or by introducing another receptionist.  

In our case, we currently have bpiSubjectAgent and bpiSubjectStorageAgent. One deals with everything BpiSubject related apart from storage - which is the job of the storage agent. As we grow, we might decide to split the bpiSubjectAgent into multiple different agents (new roles) which would help us keep the amount of responsibilities in a role on a level that is easy to read and maintain. As the user load grows, we might decide to introduce a  pub\sub bus were multiple agents of the same kind (new persons) listen to events and process them (microservices).

A command handler can inject multiple different agents in order to orchestrate the execution of  the command at hand. It is tasked with declaring WHAT needs to happen (thus agent method names should be highly descriptive) and in what order.

The agents are tasked with HOW. Agents can share among themselves helper services they might need (think of them as tools - both cashier and receptionist rely on the same Excel sheet to give the guest a preliminary price calc).

Because of the fact that you could have multiple agents in a single command who are all part of a single business transaction, we need to treat the storage role as an agent so that we can commit transaction as a last step of the execution of the command. Otherwise storage could be abstracted away behind an agent.

examples/bri-3/src/bri/identity/bpiSubjects/models/bpiSubject.ts
ognjenkurtic: Because this is just an example. Usually you will update more than one field as part of the transaction (i.e. UpdateGeneralInfomation). Updating the fields will be accompanied by a set of business rules, and we want to avoid the possibility that somewhere you can introduce a change directly and elsewhere you do it through a method. And we are good OO pupils ;).

examples/bri-3/src/bri/identity/bpiSubjects/persistence/bpiSubjects.repository.ts
ognjenkurtic: the method is async. This is similar to Task return type in net core.

examples/bri-3/test/bri.postman_collection.json
ognjenkurtic: Fixed

examples/bri-3/src/bri/identity/bpiSubjects/persistence/bpiSubjects.repository.ts
skosito: ah ok, i was just using async in js context, makes sense

skosito (dismissed):
Thanks for addressing all comments, approved

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/updateBpiSubject/updateBpiSubjectCommand.handler.ts
Therecanbeonlyone1969: > Because of the fact that you could have multiple agents in a single command who are all part of a single business 
> transaction, we need to treat the storage role as an agent so that we can commit transaction as a last step of the 
> execution of the command. Otherwise storage could be abstracted away behind an agent.

@ognjenkurtic each agent in a command might need storage as a tool. so should we have storage as BOTH agent and a helper service I wonder? cc @skosito

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
```diff
@@ -1,22 +1,52 @@
...
 
   public createNewExternalBpiSubject(name :string, description: string, publicKey: string): BpiSubject {
     return new BpiSubject(uuid(), name, description, BpiSubjectType.External, publicKey );
   }
+
+  public async fetchUpdateCandidateAndThrowIfUpdateValidationFails(id: string, name: string, desc: string, pk: string): Promise<BpiSubject> {
+    const bpiSubjectToUpdate = await this.repo.getBpiSubjectById(id);
+
+    if(!bpiSubjectToUpdate) {
+      throw new NotFoundException(`Bpi Subject with id: ${id} does not exist.`)
```
Therecanbeonlyone1969, line 30: ok a nitpick but leading to a more general point -> should we not have an error/notification message file with the messages in it in case we need to use them more than once? cc @skosito @Kasshern @Manik-Jain @Ybittan

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
```diff
@@ -1,22 +1,52 @@
...
+    }
+
+    return bpiSubjectToUpdate;
+  }
+
+  public updateBpiSubject(bpiSubjectToUpdate: BpiSubject, name :string, description: string, publicKey: string) {
```
Therecanbeonlyone1969, line 36: the non-empty validation is done in the class definition, right?

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
```diff
@@ -1,22 +1,52 @@
...
+    bpiSubjectToUpdate.updateDescription(description);
+    bpiSubjectToUpdate.updatePublicKey(publicKey);
+  }
+
+  public async fetchDeleteCandidateAndThrowIfDeleteValidationFails(id: string): Promise<BpiSubject> {
+
+    const bpiSubjectToDelete = await this.repo.getBpiSubjectById(id);
+
+    if(!bpiSubjectToDelete) {
+      throw new NotFoundException(`Bpi Subject with id: ${id} does not exist.`)
```
Therecanbeonlyone1969, line 47: see my comment above.

Therecanbeonlyone1969 (commented):
Ready to approve modulo the message comment I left

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/updateBpiSubject/updateBpiSubjectCommand.handler.ts
ognjenkurtic: I think of tools usually as pure functions - they do not produce side effects. In our domain, a proper tool is for example a class (service) that is responsible for performing signature verification. Input from the caller -> Verification -> Output to the caller - all without changing the global state of the application. When thinking about storage in this light it is clear that it is not a tool - it deals with a cross-cutting concern (DB transaction). 

If all agents are allowed to perform storage, you as a dev would need to actually read the implementation of every agent method used in the command to make sure that the result of the business transaction is properly stored or rejected. This would quickly turn into a mess to maintain and extend as we develop further.

If, on the other hand, only the storageAgent performs the commit of the transaction at the end of the command execution, this tracking of the transaction becomes trivial. The command handler which describes the WHAT and the order of execution becomes the only entity that is responsible for accepting or rejecting the storage of the result of of work of all the previous agents, based on the  validity of the business transaction.

If there is a need to execute multiple DB transactions as part of a single API call, this is usually a sign that you should split the command into multiple commands and chain them. This is what is called Saga in NestJS. 

I suggest we continue with this approach - as we get to more complex business cases, the usage and the motivation behind the architecture will become more clear. We can always adjust and refactor as we go, if the need arises.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
ognjenkurtic: Yes, good point. I ll introduce constants for now and we can later think of internalization.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
ognjenkurtic: Yes. We will have one level of protection (i.e. from external inputs) through input DTO validation (#520) and another level of protection (i.e. from accidental bugs introduced by devs) in the domain object. Finally the DB will reject the update if the constrainst are not followed.

For point 3 we ll introduce global exception handling in the upcoming days.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
ognjenkurtic: I hope i answered above.

ognjenkurtic:
image

Here's a picture of a liftboy for those who are better at grasping new concepts visually.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
ognjenkurtic: Added a primitive mechanism for shared err messages. Let's build on it further as we go.

Therecanbeonlyone1969 (approved):
LGTM

examples/bri-3/src/bri/identity/bpiSubjects/api/err.messages.ts
```diff
@@ -0,0 +1,2 @@
+export const NOT_FOUND_ERR_MESSAGE = "Bpi Subject does not exist.";
+export const NAME_EMPTY_ERR_MESSAGE = "Name cannot be empty.";
```
Kasshern, line 2: This error message feels a little unspecific, but I approve anyways.

Kasshern (approved):

Issue #482, opened 2022-07-12
SRI Storage Component - Database Preparation Tasks


Ybittan:
## Overview

It has been determined that the SRI will need a database. Postgres is the current system of choice. 
This issue will outline the initial tasks necessary to spin up a DB.

### Reference

- [SRI Storage](https://github.com/eea-oasis/baseline/issues/478)

### Acceptance

- Should spin up a database on the machine's local server and connect to it
- Should have a plan for future state of a containerized (Docker) db to spin up fresh

## Tasks

- [x] Create db programmatically
- [x] Use migration library to connect to the DB
- [x] Plan for future Docker components
- [x] Spin up Docker postgres container and connect to it

ognjenkurtic:
Please research tools to enable easy migrations of the db as we introduce new functionalities and models.

Ybittan:
[What data must be stored document](https://docs.google.com/document/d/1RoK9kDxvL4Opg9SSmgdCpgbK88zIJMBJv0umRtR6AZ0/edit?usp=sharing)

Ybittan:
Updated based on suggestions from a comment, reference below:

"I do not think you should and could map all the necessary data at this point.For the purpose of issue https://github.com/eea-oasis/baseline/issues/482, i think it is enough to map the general requirements from the storage layer (as you already did bellow), and then decide on the storage tech itself as well as the client to be used to communicate with it. As part of https://github.com/eea-oasis/baseline/issues/483, you can then implement the client with these basic functionalities setup (i.e. migrations, CRUD patterns...) and every dev will implement the necessary migrations for the scope of the issue he\she is working on." -Ogjnen

ognjenkurtic:
Hey @Ybittan , this one might be interesting for us https://typeorm.io/

PR #522, opened 2022-09-15
Feature/482 sri storage sql calls

Ybittan:
# Description

This PR spins up a pg docker image, and connects to it.
It also provides example code for a local db, and example queries to execute. 

## Related Issue
https://github.com/eea-oasis/baseline/issues/482

## Motivation and Context
 Introduces the necessary storage components requested in the issue

## How Has This Been Tested

Manually xecuted queries against local db and docker db several times to ensure good connection and expected behavior. 

## Types of changes

## Checklist


examples/bri-3/components/storage/sql-connection.js
```diff
@@ -0,0 +1,79 @@
...
+// const client = new Client({
+//     user: 'postgres',
+//     host: 'localhost',
+//     database: 'postgres',
+//     password: '',
+//     port: 5432,
+// });
+
+//Docker DB
+const client = new Client({
```
skosito, line 16: multiple db environments should be in multiple env files, not sure what is storate/database.json used for?

examples/bri-3/components/storage/sql-connection.js
```diff
@@ -0,0 +1,79 @@
...
+    user: 'postgres',
+    host: 'localhost',
+    database: 'postgres',
+    password: 'example',
+    port: 5433,
+});
+
+client.connect();
+
+const ordersTable = `CREATE TABLE orders (
```
skosito, line 26: all these table queries should be part of init migration

examples/bri-3/components/storage/sql-connection.js
skosito, line 16: not sure about organization of storage modules but if this module is just for connection client or method for creating client should be exported from this module

examples/bri-3/components/storage/sql-connection.js
```diff
@@ -0,0 +1,79 @@
...
+const stateTable = `CREATE TABLE state (
+    id int, --[pk, increment]
+    agreementState varchar,
+    acceptanceStatus varchar
+);`;
+
+const tables = [ordersTable, bpiTable, workstepsTable, bpiTimestampRecordsTable, stateTable]
+
+//run a query to run the create tables listed above, included in the tables array
+tables.forEach(table => {
```
skosito, line 61: when init migration is there this is redundant

examples/bri-3/components/storage/migrations/20220818160540-mydb.js
```diff
@@ -0,0 +1,27 @@
...
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function(db) {
+  return null;
```
skosito, line 18: dummy migration is not useful, just for illustration, but since we can create tables lets make proper migration
also there is migration introduced by @ognjenkurtic already i believe, let's do same approach/folders for it

examples/bri-3/components/storage/database.json
```diff
@@ -0,0 +1,11 @@
+{
```
skosito, line 1: this file looks redundant to me

Ybittan: Agreed, this was for the local db, before implementing Docker. This is not necessary anymore, I just forgot to remove it.

examples/bri-3/components/storage/sql-connection.js
Therecanbeonlyone1969: should the client configuration not come from a config file ... e.g. clientDBConfig.json?

examples/bri-3/components/storage/migrations/20220818160540-mydb.js
ognjenkurtic: This file is also not needed. Migrations are handled through prisma commands for local development and will be triggered from the app on application startup in the future.

examples/bri-3/components/storage/package.json
```diff
@@ -0,0 +1,15 @@
+{
```
ognjenkurtic, line 1: No need to have a separate package.json in the storage component. BRI-3 will deal with migrations on app startup.

examples/bri-3/components/storage/sql-connection.js
ognjenkurtic: Yes, all the entities will be added through  prisma.schema and then migrations (or a init migration) generated though prisma commands. This migration\s will then be executed by the app on startup. This file is not needed.

ognjenkurtic (commented):
I think we have a misunderstanding here. All migrations will be generated through prisma, by editing the prisma.schema and running prisma generate. These migrations will be executed on app startup (or by prisma command locally). 

Regarding postgres, we just need a set of commands added to readme somewhere that explain how to fire up a postgres container as well as what connection params to place in the .env in order to be able to connect to it.

examples/bri-3/components/storage/sql-connection.js
Manik-Jain: +1

examples/bri-3/components/storage/sql-connection.js
Manik-Jain: Agreed. @ognjenkurtic could you please validate if the default ORM setting to "always" spin-up tables is disabled?

ognjenkurtic: It is not setup atm to do anything when the app starts. We manually trigger db upgrades with prisma cli.

examples/bri-3/src/bri/storage/client-congif.json
```diff
@@ -0,0 +1,9 @@
+{
```
ognjenkurtic, line 1: i believe this is not needed. what is the use of this file? it also has a typo in filename, should be con**fig**

examples/bri-3/prisma/schema.prisma
```diff
@@ -16,4 +16,36 @@ model BpiSubject {
   description String  
   type Int
   publicKey String 
+}
+
+model Order {
```
ognjenkurtic, line 21: You can remove all of these for now. They will be added in interations as each component is introduced to the codebase

ognjenkurtic (approved):
Please address the two comments and it is ready to merge in my eyes

examples/bri-3/.env sample
```diff
@@ -4,4 +4,4 @@
 # Prisma supports the native connection string format for PostgreSQL, MySQL, SQLite, SQL Server, MongoDB and CockroachDB.
 # See the documentation for all the connection string options: https://pris.ly/d/connection-strings
 
-DATABASE_URL="file:./dev.db"
```
skosito, line 7: this file should be named `.env.sample` not `.env sample`

please add separate `.env` file and git ignore it, `.env.sample` is just used to show which variables need to be set for local development and `.env` will be set up by everyone locally

also usual postgres port is 5432

examples/bri-3/src/bri/storage/client-congif.json
skosito: yea not needed, connection string from env variable should be enough, please remove

examples/bri-3/src/bri/storage/readme.md
```diff
@@ -0,0 +1,5 @@
+Use the following console command to spin up the docker db
```
skosito, line 1: lets use usual postgres port 5432:5432
also we can add here  in readme that .env  connection string should be set with these params

skosito (approved):
left some comments to be addressed before merge, but they are simple changes so approved

examples/bri-3/src/bri/storage/readme.md
Ybittan: 5433 is to prevent a local db already running on that port from interfering.

examples/bri-3/.env sample
Ybittan: 5433 is to prevent a local db already running on that port from interfering.

Ybittan: I have my own .env which is ignored. I added to the sample file so that other devs have an example string for spinning up docker in their own .env file.

examples/bri-3/src/bri/storage/readme.md
skosito: not sure i follow, this is local db, this project doesn't require 2 local dbs running, locally everybody can set port that works in .env file but for illustration we should use 5432 since that is default postgres port in all docs... for example, maybe i have 5 dbs container running for other projects so 5433 wont work either

 when we have alice and bob dbs we will then run 2 docker containers, one on 5432 second on 5433 for example, but now we have only one

examples/bri-3/.env sample
Ybittan: Okay, I can make that change. I did not create this sample file, only edited it.

skosito (approved):

ognjenkurtic (approved):

Kasshern (approved):

Issue #493, opened 2022-08-02
Introduce basic CI\CD for SRI


ognjenkurtic:
## Overview

We need a basic CI|CD pipeline for the SRI project

### Reference

- [Follows from this](https://github.com/eea-oasis/baseline/issues/491)
- [And then this](https://github.com/eea-oasis/baseline/issues/492)

### Assumptions

- Github actions
- Deployments come later

### Acceptance

 - CI pipeline setup and running linting on every merge into main

## Tasks

- [ ] Introduce Github Action
- [ ] Trigger project build and lint

PR #524, opened 2022-09-15
Feature/492 linter, prettier and github actions

skosito:
# Description

This PR adds following:
- new scripts to perform prettier and linter  checks
- run prettier formatting fixes - **this is why there are 53 files changed, but most of these are just automatic formatting fixes**
- add tabWidth 2 to prettier config, the rest of linter and prettier config is left as nest js default
- add github workflows to run prettier and linter checks and test on every PR to main and push to main only when files in bri-3 folder are changed
- all lint warnings and errors are fixed

## Related Issue

https://github.com/eea-oasis/baseline/issues/492
https://github.com/eea-oasis/baseline/issues/493

## Motivation and Context

Check issues for more context

## How Has This Been Tested

Linter and prettier locally, github actions on my fork:  https://github.com/skosito/baseline/pulls
Here I have 2 PRs, one changing files in bri-3, triggering github action with every commit on PR, and one outside bri-3 folder not triggering any actions. Action runs can be seen here: https://github.com/skosito/baseline/actions

Also, github actions running on this PR too https://github.com/eea-oasis/baseline/actions/runs/3063436398

## Types of changes

## Checklist


.github/workflows/bri-3.yml
```diff
@@ -0,0 +1,55 @@
...
+  push:
+    paths:
+      - examples/bri-3/**
+    branches:
+      - main
+  pull_request:
+    paths:
+      - examples/bri-3/**
+    branches:
+      - main
```
ognjenkurtic, line 13: this means it executes on all PRs to main?

examples/bri-3/src/bri/identity/bpiAccounts/models/bpiAccount.ts
```diff
@@ -1,11 +1,11 @@
 export class BpiAccount {
```
ognjenkurtic, line 1: could you please execute this after all the PRs are merged, and then recreate your PR with latest changes? Otherwise we would need to put a lot of effort into merging.

examples/bri-3/.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
   "singleQuote": true,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "tabWidth": 2
```
Manik-Jain, line 4: AFAIK, this is the default setting. Recommended setting is 4 or higher. Should we not not increase the vaalue here?

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
```diff
@@ -1,22 +1,36 @@
...
-import { uuid } from 'uuidv4'; 
- 
 @Injectable()
 export class BpiSubjectAgent {
-  // Agent methods have extremely declarative names and perform a single task 
-  public throwIfCreateBpiSubjectInputInvalid(name: string, desc: string, pk: string) {
+  // Agent methods have extremely declarative names and perform a single task
+  public throwIfCreateBpiSubjectInputInvalid(
+    name: string,
+    desc: string,
```
Manik-Jain, line 12: I think it would be a small refactoring task to rename desc -> description, and pk -> publicKey. Can we push in this change?

Manik-Jain (approved):

examples/bri-3/src/bri/identity/bpiAccounts/models/bpiAccount.ts
skosito: yes i will do this, in this PR after all pending now is merged

.github/workflows/bri-3.yml
skosito: on all PRs on main that change files in bri-3 folder, i opened PR on my fork (url in PR description) that modifies other files and actions were not executed
path regex is set both on workflow level (paths), and on job level (working-directory)

examples/bri-3/.prettierrc
skosito: you are right, default is 2 i will remove this, i put 2 because files currently in project all had that width
i am ok with any tab settings as long as everybody agrees, if there is no objection on PR i will just remove this and we will use default value 2

ognjenkurtic: You meant default value **4** i guess.

ognjenkurtic (approved):

examples/bri-3/.prettierrc
skosito: no default is 2, i think leaving  2 explicitly is better, it is same result with this and without this

ognjenkurtic (approved):

Kasshern (approved):

Issue #517, opened 2022-09-14
Write generic mappers for BRI-3 (SRI)


skosito:
## Overview

Check following TODOs:
- Write generic mapper domainObject -> DTO
- Write generic mapper prismaModel -> domainObject
- Write generic mapper domainObject -> prismaModel

### Reference

TODOs introduced in this PR https://github.com/eea-oasis/baseline/pull/515

ognjenkurtic:
Added additional TODO comments related to mapping to the description of the issue.

PR #528, opened 2022-09-25
517 generic mapper

Manik-Jain:
# Description

Adding a Generic mapper to convert object to Type at runtime

## Related Issue

https://github.com/eea-oasis/baseline/issues/517

## Motivation and Context

We need a general purpose mapper to convert object to Type at runtime, with capabilities to define some explicit properties on the type

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist

- [X] My change requires a change to the documentation.


examples/bri-3/package.json
```diff
@@ -34,7 +34,11 @@
     "ew-precise-proofs-js": "^1.2.0",
     "reflect-metadata": "^0.1.13",
     "rimraf": "^3.0.2",
-    "rxjs": "^7.2.0"
+    "rxjs": "^7.2.0",
+    "ts-node-dev": "^2.0.0",
+    "tst-reflect": "^0.12.0",
+    "tst-reflect-transformer": "^0.12.1",
+    "uuidv4": "^6.2.13"
```
ognjenkurtic, line 41: I believe we removed this one previously.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjectsStorage.agent.ts
```diff
@@ -10,73 +12,41 @@ import { BpiSubject } from '../models/bpiSubject';
...
-        bp.type,
-        bp.publicKey,
-      );
+    return bpiSubjectModels.map((bpiSubjectModel) => {
+      return Mapper.map(bpiSubjectModel, getType<BpiSubject>()) as BpiSubject;
     });
   }
 
   async createNewBpiSubject(bpiSubject: BpiSubject): Promise<BpiSubject> {
+    const { name, description, publicKey, type } = bpiSubject;
```
ognjenkurtic, line 32: why introducing this additional line when the bpiSubject is already containing the params. This is especially cumbersome when there are many params.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjectsStorage.agent.ts
```diff
@@ -10,73 +12,41 @@ import { BpiSubject } from '../models/bpiSubject';
...
     });
-
-    return new BpiSubject(
-      newBpiSubjectModel.id,
-      newBpiSubjectModel.name,
-      newBpiSubjectModel.description,
-      newBpiSubjectModel.type,
-      newBpiSubjectModel.publicKey,
-    );
+    return Mapper.map(
```
ognjenkurtic, line 46: one liner?

examples/bri-3/src/bri/utils/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @author : Manik-Jain
+ */
+export default class Mapper {
+  /**
+   * type convert from provided input to provided Type
+   * provides capabilities to define some explicit
+   * properties on the type
+   *
+   * Be careful with providing opts obeject as
```
ognjenkurtic, line 15: do we need opts at all?
ob**e**ject

examples/bri-3/src/bri/utils/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+        Object.keys(opts).forEach((prop) => {
+          if (keys.includes(prop)) {
+            result[prop] = opts[prop];
+          } else {
+            mismatchedKeys.push(prop);
+          }
+        });
+      }
+
+      console.log(`Type conversion successful`);
```
ognjenkurtic, line 48: no need to log success as there will be a lot of them

examples/bri-3/src/bri/utils/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+      if (mismatchedKeys.length > 0) {
+        console.log(
```
ognjenkurtic, line 50: we should consider creating mapper as a singleton service so that it can be dependency injected via nestjs and so that  we can inject logger later.

examples/bri-3/src/bri/utils/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+          `Found ${mismatchedKeys.length} mismatched keys => ${mismatchedKeys}`,
+        );
+      }
+
+      return result;
+    } else if (classType.kind === TypeKind.Class) {
+      const constructorParams = classType.getConstructors()[0].getParameters();
```
ognjenkurtic, line 57: Why are we working with constructors params in case Class? Why not repeat the same thing as with TypeKind.Interface? Why only mandatoryParams?

examples/bri-3/src/bri/utils/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+          if (classProps.includes(prop)) {
+            result[prop] = opts[prop];
+          } else {
+            mismatchedKeys.push(prop);
+          }
+        });
+      }
+    }
+
+    console.log(`Type conversion successful`);
```
ognjenkurtic, line 92: no need

examples/bri-3/src/bri/utils/test.ts
```diff
@@ -0,0 +1,14 @@
+// import Mapper from './mapper';
+// import { getType, Type, PropertyInfo } from "tst-reflect";
+// import { BpiSubjectDto } from './bpiSubject.dto';
+
+// const from  = {
```
ognjenkurtic, line 5: please make sure to add a few tests

examples/bri-3/src/bri/workgroup/worksteps/agents/workstepsStorage.agent.ts
```diff
@@ -2,92 +2,79 @@ import { Injectable, NotFoundException } from '@nestjs/common';
...
   async createNewWorkstep(workstep: Workstep): Promise<Workstep> {
+    const {
+      id,
+      name,
+      version,
+      status,
+      workgroupId,
+      securityPolicy,
+      privacyPolicy,
+    } = workstep;
```
ognjenkurtic, line 39: workstep is already containing all the params, no need to introduce these new lines and reduce readability.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjectsStorage.agent.ts
Manik-Jain: I'm going to be replacing this whole piece with one liner. Some more code coming in next commit

examples/bri-3/src/bri/utils/access-modifiers.ts
```diff
@@ -0,0 +1,5 @@
+export declare enum AccessModifier {
```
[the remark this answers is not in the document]
Manik-Jain: Yes, this can be done. Nice idea actually

examples/bri-3/src/bri/utils/mapper.ts
Manik-Jain: As mentioned in the JSDoc, this is for future proofing to handle scenarios when we want to set an explicit value on the mapped object. A minimalistic example can be setting subject type to external.

examples/bri-3/src/bri/utils/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+   *
+   * Be careful with providing opts obeject as
+   * this will change the properties on the type
+   *
+   * @param from
+   * @param classType
+   * @param opts
+   * @returns generated object
+   */
+  static map<T>(from: any, classType: Type, opts?: any) {
```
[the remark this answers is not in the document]
Manik-Jain: Right. I'll remove this, I was actually using generics before the ttsc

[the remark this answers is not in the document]
Manik-Jain: probably, I'll rename this to objectType to signify class as well as interface types

examples/bri-3/src/bri/utils/mapper.ts
Manik-Jain: I think this will be helpful, specifically in debugging scenarios.

examples/bri-3/src/bri/utils/mapper.ts
Manik-Jain: As you know we can and we have constructors defined only in Classes. The idea is, 

for class type : initialize the object with only the parameters used in the  defined constructor, which is, in code, is being referred as `mandatoryParams`

for interface type : initialize the object with all the mandatory params, as we can have some optional params too in the interface.

examples/bri-3/src/bri/workgroup/worksteps/agents/workstepsStorage.agent.ts
Manik-Jain: will be replaced with one-liner in next commit

examples/bri-3/src/bri/utils/mapper.ts
Manik-Jain: change done

examples/bri-3/src/bri/utils/test.ts
Manik-Jain: will add more tests after this is figured out : https://github.com/eea-oasis/baseline/issues/534

ognjenkurtic:
Hey @Manik-Jain  , what is the status with this one?  I believe it is not mergeable until the issue leading to commented tests is  resolved. Do you think it would make sense that  we fall back to a simple mapper doing something like object.assign while  we wait?

Therecanbeonlyone1969:
> Hey @Manik-Jain , what is the status with this one? I believe it is not mergeable until the issue leading to commented tests is resolved. Do you think it would make sense that we fall back to a simple mapper doing something like object.assign while we wait?

I agree. Let's keep things simple and moving. We can always come back to this in the next iteration @Manik-Jain

ognjenkurtic:
> @Therecanbeonlyone1969 Yeah, we can circle back to this PR once the repository is properly configured. I see that the `test` folder provided by Nest hasn't been utilized for the purpose of adding unit test suites, which is why the compilation isn't adding anything to `dist` directory. This needs to be
> 
> @ognjenkurtic I gave it a thought, and we can use Object.assign at the cost of type-checking. This can be used temporarily while I fix the test case configuration

@Manik-Jain Repository is properly configured when it comes to testing. It is following a common pattern for unit and e2e tests. From [NestJs documentation](https://docs.nestjs.com/fundamentals/testing):

Unit tests:
_Keep your test files located near the classes they test. Testing files should have a .spec or .test suffix._

E2E tests:
_Keep your e2e test files inside the test directory. The testing files should have a .e2e-spec suffix._


TLDR; Test directory is meant for E2E tests as they are testing the whole scope of the application. Unit tests are next to the things they are testing as this is the scope they are interested in. I would suggest we continue using this approach unless there is some exceptional thing we need to consider.

re: Object.assign. Great, do you want to do that as part of another PR? If yes, please feel free to open an issue.

Manik-Jain:
> > @Therecanbeonlyone1969 Yeah, we can circle back to this PR once the repository is properly configured. I see that the `test` folder provided by Nest hasn't been utilized for the purpose of adding unit test suites, which is why the compilation isn't adding anything to `dist` directory. This needs to be
> > @ognjenkurtic I gave it a thought, and we can use Object.assign at the cost of type-checking. This can be used temporarily while I fix the test case configuration
> 
> @Manik-Jain Repository is properly configured when it comes to testing. It is following a common pattern for unit and e2e tests. From [NestJs documentation](https://docs.nestjs.com/fundamentals/testing):
> 
> Unit tests: _Keep your test files located near the classes they test. Testing files should have a .spec or .test suffix._
> 
> E2E tests: _Keep your e2e test files inside the test directory. The testing files should have a .e2e-spec suffix._
> 
> TLDR; Test directory is meant for E2E tests as they are testing the whole scope of the application. Unit tests are next to the things they are testing as this is the scope they are interested in. I would suggest we continue using this approach unless there is some exceptional thing we need to consider.
> 
> re: Object.assign. Great, do you want to do that as part of another PR? If yes, please feel free to open an issue.

Hi @ognjenkurtic  @Therecanbeonlyone1969 

As discussed during the weekly standup, I made a final attempt to fix the failing test cases with reflect TS compiler, and the test cases are all passing now.
So, we don't need to move to a new approach as this one works with proper type-safety.

Thanks

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjectsStorage.agent.ts
```diff
@@ -2,81 +2,65 @@ import { Injectable, NotFoundException } from '@nestjs/common';
...
     const bpiSubjectModels = await this.bpiSubject.findMany();
-    return bpiSubjectModels.map((bp) => {
-      return new BpiSubject(
-        bp.id,
-        bp.name,
-        bp.description,
-        bp.type,
-        bp.publicKey,
-      );
+    console.log(bpiSubjectModels);
```
ognjenkurtic, line 33: Please remove console.log

examples/bri-3/src/bri/identity/bpiSubjects/agents/mockBpiSubjectStorage.agent.ts
```diff
@@ -29,16 +38,22 @@ export class MockBpiSubjectStorageAgent {
...
 
     this.bpiSubjectsStore.push(createdBp);
-
     return Promise.resolve(createdBp);
   }
 
   async updateBpiSubject(bpiSubject: BpiSubject): Promise<BpiSubject> {
-    const bpToUpdate = this.bpiSubjectsStore.find(
+    const bpToUpdate = this.mapper.map(
+      this.bpiSubjectsStore.find((bp) => bp.id === bpiSubject.id),
```
ognjenkurtic, line 46: There is a null reference bug here in case this.bpiSubjectStore.find does not return a bpiSubject.

This whole method is strange. We find by id and map to BpiSubject something which is already BpiSubject.
Then we find the index of the same object and use ti to store the changes. I believe we can ditch lines 45, to 49 completely or am i  missing something?

examples/bri-3/src/bri/identity/bpiSubjects/api/subjects.controller.spec.ts
```diff
@@ -102,15 +104,16 @@ describe('SubjectController', () => {
       // Act
       const bpiSubjects = await sController.getAllBpiSubjects();
 
-      // Assert
+      //TODO : [MJ] fix this with ttsc
```
ognjenkurtic, line 107: comment is obsolete?

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/createBpiSubject/createBpiSubjectCommand.handler.ts
```diff
@@ -8,8 +8,8 @@ export class CreateBpiSubjectCommandHandler
   implements ICommandHandler<CreateBpiSubjectCommand>
 {
   constructor(
-    private agent: BpiSubjectAgent,
-    private storageAgent: BpiSubjectStorageAgent,
+    private readonly agent: BpiSubjectAgent,
+    private readonly repo: BpiSubjectStorageAgent,
```
ognjenkurtic, line 12: name should be storageAgent.

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/updateBpiSubject/updateBpiSubjectCommand.handler.ts
```diff
@@ -13,18 +14,11 @@ export class UpdateBpiSubjectCommandHandler
...
-
-    this.agent.updateBpiSubject(
-      bpiSubjectToUpdate,
-      command.name,
-      command.desc,
-      command.publicKey,
-    );
-
-    await this.storageAgent.updateBpiSubject(bpiSubjectToUpdate);
+    Object.assign(bpiSubjectToUpdate, command);
```
ognjenkurtic, line 21: Why is agent method removed and object assign used instead? The goal is to avoid HOW on the level on the command and instead just explain WHAT through declarative agent method. It is the job of the agent method to deal with the actual implemenentation (which is simple object assign for now but will be something more complicated as we move forward.)

examples/bri-3/src/bri/utils/mapper.ts
```diff
@@ -0,0 +1,134 @@
...
+ * Generic mapper to convert object to Type at runtime
+ *
+ * @author : Manik-Jain
+ */
+@Injectable()
+export default class Mapper {
+  /**
+   * type convert from provided input to provided Type
+   * provides capabilities to define some explicit
+   * keyerties on the type
```
ognjenkurtic, line 16: properties

examples/bri-3/src/bri/utils/mapper.ts
```diff
@@ -0,0 +1,134 @@
...
+   *
+   * Be careful with providing options object as
+   * this will change the keyerties on the type
```
ognjenkurtic, line 19: properties

examples/bri-3/test/jest-e2e.json
```diff
@@ -1,9 +1,33 @@
 {
-  "moduleFileExtensions": ["js", "json", "ts"],
-  "rootDir": ".",
-  "testEnvironment": "node",
-  "testRegex": ".e2e-spec.ts$",
-  "transform": {
-    "^.+\\.(t|j)s$": "ts-jest"
-  }
+	"compilerOptions": {
```
ognjenkurtic, line 2: i am just curious why was there a need to touch e2e tests configuration?

examples/bri-3/src/bri/utils/mapper.ts
```diff
@@ -0,0 +1,134 @@
...
+          );
+        }
+      }
+
+      return result;
+    } else if (object.kind === TypeKind.Class) {
+      const constructorParams = object.getConstructors()[0].getParameters();
+      const classProperties = object
+        .getProperties()
+        .filter((property) => property.accessModifier === 2)
```
ognjenkurtic, line 70: what is 2? Could it be moved to a constant at the top of the file or exchanged for an enum?

examples/bri-3/src/bri/identity/bpiSubjects/agents/mockBpiSubjectStorage.agent.ts
Manik-Jain: No, there is nothing strange happening here. Since we are storing the mock subjects in an array, we need to update the object with the exact value as the received input so that the getter returns the updated value instead of what was there before the update. 

Also, BpiSubject class has variables as {_id, _name,...} so the returned object needs to be mapped to the corresponding {id, name,...} in order for the Object.assign to work.

examples/bri-3/test/jest-e2e.json
Manik-Jain: this is done to future proof when we start adding integration tests.

examples/bri-3/src/bri/identity/bpiSubjects/agents/mockBpiSubjectStorage.agent.ts
ognjenkurtic: I am not sure i understand. Why would not this work:

`const bpToUpdate = this.bpiSubjectsStore.find(
      (bp) => bp.id === bpiSubject.id,
    );`

`Object.assign(bpToUpdate, bpiSubject);`

`return Promise.resolve(bpToUpdate);`

Both bpToUpdate and BpiSubject are of same type since we are mocking the storage layer.

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/updateBpiSubject/updateBpiSubjectCommand.handler.ts
Manik-Jain: this method is calling update name, description, publicKey methods on BpiSubject which are doing essentially the same thing as Object.assign

examples/bri-3/src/bri/identity/bpiSubjects/agents/mockBpiSubjectStorage.agent.ts
Manik-Jain: The point that I'm trying to make here is:

The test case has an update call and a getter call. 
As you mentioned, update method returns the Object.assign method.. but it doesn't update the mocked array.
So, the array is still holding the unassigned object.

Manik-Jain: @ognjenkurtic  Actually I gave it a deeper thought, and the reason is the difference between the input and the create BpiSubject keys. Input is {id, name,...} and what's in the array is {_id, _name,...}
I think we can live with this for the test case scenario.

ognjenkurtic: Is this becuase we are not mapping from entity mode (Prisma generated model) to our domain object (BpiSubject class) during the fetching from the db? There should always be a mapping in both directions on the level of the storage agent so that the rest of the app talks only the language of our own domain objects.

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/updateBpiSubject/updateBpiSubjectCommand.handler.ts
ognjenkurtic: The idea is that the state of a domain object is always changed through it's methods. They are simple now, but will become much more complex as we move forward. Also, we introduced the agent approach to make sure that the abstraction level is always the same on the level of the command. You just read the names of the methods and you understand what is happening. HOW is hidden away inside of the agent method, which can further change the state of the domain object through the  exposed methods.

Manik-Jain: as agreed, code changes have been pushed

examples/bri-3/src/bri/identity/bpiSubjects/agents/mockBpiSubjectStorage.agent.ts
Manik-Jain: as agreed during the call, the mock generator has been fixed

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/updateBpiSubject/updateBpiSubjectCommand.handler.ts
```diff
@@ -14,11 +14,26 @@ export class UpdateBpiSubjectCommandHandler
...
 
   async execute(command: UpdateBpiSubjectCommand) {
-    const bpiSubjectToUpdate: BpiSubject =
+    const bpiSubject: BpiSubject =
       await this.agent.fetchUpdateCandidateAndThrowIfUpdateValidationFails(
         command.id,
       );
-    Object.assign(bpiSubjectToUpdate, command);
+
+    const bpiSubjectToUpdate = new BpiSubject(
```
ognjenkurtic: This mapping already inside of fetchUpdateCandidateAndThrowIfUpdateValidationFails or am i missing something?

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
```diff
@@ -42,13 +42,20 @@ export class BpiSubjectAgent {
...
-    const bpiSubjectToUpdate = await this.storageAgent.getBpiSubjectById(id);
+    const bpiSubjectToUpdate: BpiSubject =
+      await this.storageAgent.getBpiSubjectById(id);
 
     if (!bpiSubjectToUpdate) {
       throw new NotFoundException(NOT_FOUND_ERR_MESSAGE);
     }
 
-    return bpiSubjectToUpdate;
+    return new BpiSubject(
```
ognjenkurtic, line 52: can we use mapper here as well?

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/updateBpiSubject/updateBpiSubjectCommand.handler.ts
Manik-Jain: Oh right, i guess I changed this already. Let me check please if I committed this

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
Manik-Jain: Mapper needs a little modification to invoke the constructors dynamically. I'll make that change in a separate PR.

examples/bri-3/src/bri/identity/bpiSubjects/capabilities/updateBpiSubject/updateBpiSubjectCommand.handler.ts
Manik-Jain: Oh, I just checked. 
I think you were checking the older commit. please check this commit: 

https://github.com/eea-oasis/baseline/pull/528/commits/6ebd72657ecef3e49135ceffebd15ea8e9b31cd2

skosito (approved):
hey @Manik-Jain i am approving with one question, i dont want to block PR since a lot of time and discussion went into it, but i googled a bit for typescript automapper and found this https://automapperts.netlify.app/docs/nestjs
automapper that has integration with nestjs, from quick skim through docs seems to be working similarly to .net automapper library, looks like it is very simple to use

just wondering why not using something like this and instead building solution from scratch including reflection, changing build and compilation process with new transformer etc?

Manik-Jain:
> https://automapperts.netlify.app/docs/nestjs

thanks for this @skosito 
This looks useful. Let me spend some time with it

Issue #521, opened 2022-09-15
Decide on logging strategy and library for BRI-3 (SRI)


skosito:
## Overview

Choose library and propose and implement strategy for logging in bri-3. Remove console.logs the we currently have.

### Acceptance

- Install logger library in bri-3
- Replace console.logs in bri-3 folder with logger library

## Tasks

- [ ] Choose and install library
- [ ] Fix console.log

Therecanbeonlyone1969:
My suggestion is a common ts logger -> https://www.npmjs.com/package/typescript-logger

And for processing and visualization, we could use [Grafana](https://github.com/grafana/grafana) and postgres or mongo for storage.

PR #529, opened 2022-09-26
Feature/521 tslog installed, TODO 521 completed

Ybittan:
# Description

To install a dev-friendly typescript logging library [tslog](https://www.npmjs.com/package/tslog) that will replace current and future console logs.

## Related Issue

#521 

## Types of changes

## Checklist


package.json
```diff
@@ -1 +1,5 @@
-{}
\ No newline at end of file
+{
```
skosito, line 1: wrong package.json, bri-3/package.json is the one to install this

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -1,4 +1,6 @@
 import { PreciseProofs } from 'ew-precise-proofs-js';
+import { Logger } from "tslog";
+const log: Logger = new Logger();
```
skosito, line 3: i dont think we should instantiate new logger instance on all places where we use it, is there a way to instantiate and configure once and import everywhere where needed?

skosito (changes requested):
Please fix prettier build, also i dont think this compiles because package is not installed on proper location. Also we should check on properly instantiating logger

package.json
Ybittan: fixed

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
Therecanbeonlyone1969: @skosito each logger must be instantiated per app afaik. Unless there is a specific setting in the config file that I am not aware of. cc @Ybittan

Therecanbeonlyone1969 (dismissed):
lgtm

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
Ybittan: Removed unused logger. See LogConfig.ts and let me know if this is what you were hoping for

Therecanbeonlyone1969 (dismissed):
LGTM

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
ognjenkurtic: I believe we should follow the NestJS modular approach here as well as logging is just another service that should be dependency injected. instead of instantiating it in a LogConfig.ts., could we create a LoggingModule that has a LoggingService registered and exported as a provider which can then be injected in any other module\agent\serveice?

ognjenkurtic: You can see how this was solved for[ winston](https://github.com/gremo/nest-winston) . Unfortunaltely ts-log [does not have this support for NestJs](https://github.com/fullstack-build/tslog/issues/89), but we can build something simple. For example:
LoggingModule would be in src/bri/shared/logging and LogggingService would be next to it, with the instantiation of the logger and interface exposing methods needed to perform the logs. LoggingService would be registered as provider in the LModule and the LoggingModule would be imported in the AppModule. This way we have a singleton logging service across the app, and we just inject where we need it.

skosito: @ognjenkurtic yes agree, log config was some random example i found quickly on google but i agree it should be separate module that can be injected around
but also keep in mind we might have some files that are just plain classes or modules outside nestjs di (is this true?) that would also need to use this logger, for example merkle tree file here, but if its a module with exported methods should be enough

skosito: if we want to be in nestjs world then why not this https://docs.nestjs.com/techniques/logger ?

ognjenkurtic: Based on the docs they dont seem too supportive to use it in production. I would stick with tslog inside of a logging module. 
We should not have a need for plain classes with business logic where logging is needed - business logic should always be in an agent or a service.

skosito: not a business logic in this case but util class, so util service i presume in nestjs?
im fine with tslog, @Ybittan please let me know if you need some help with setting this up as @ognjenkurtic described

ognjenkurtic: Yes, @Ybittan please ping if you need any help with this.

Manik-Jain (dismissed):

Ybittan:
@ognjenkurtic @skosito switched from config file to nestjs module. Please review and give feedback if needed

examples/bri-3/LogConfig.ts
```diff
@@ -0,0 +1,7 @@
+import { Logger } from 'tslog';
```
ognjenkurtic, line 1: this file is now obsolete and can be removed completely

examples/bri-3/Logger.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common';
```
ognjenkurtic, line 1: this file is now obsolete and can be removed completely

examples/bri-3/src/bri/zkp/components/merkleTree/merkleTree.ts
```diff
@@ -1,5 +1,7 @@
 import { PreciseProofs } from 'ew-precise-proofs-js';
+import { getLogger } from '../../../../../LogConfig';
```
ognjenkurtic, line 2: rem these samples please, we ll also refactor this merkletree to use the nestjs modules approach

Ybittan: Updated

examples/bri-3/Logger.module.ts
Ybittan: Updated

examples/bri-3/LogConfig.ts
Ybittan: Updated

examples/bri-3/LogConfig.ts
```diff
@@ -0,0 +1,7 @@
...
+
```
ognjenkurtic, line 2: @Ybittan i think you missed to delete the files. Please recheck.

ognjenkurtic (approved):

PR #537, opened 2022-10-05
532 bri3 transactions read api

ognjenkurtic:
# Description

Adds read by id functionality to the transaction API from PR #533 .Should be reviewed after that one is merged.
Didnt add get all functionality as we still do not have from and to fields so it does not make much sense to read all transactions from everyone. 
Tests are coming in a final, separate PR.

## Related Issue

#532 

## How Has This Been Tested

Manual testing through the provided Postman collection

## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Request to be added as a Code Owner/Maintainer

## Checklist

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [X] All new and existing tests passed.
- [X] I commit to abide by the Responsibilities of Code Owners/Maintainers.

Issue #532, opened 2022-09-29
SRI Transactions - Model and CRUD Api


ognjenkurtic:
## Overview

Create SRI Transaction model and CRUD APi in the typescript language utilizing the Subjects and Accounts Component Requirements document as an outline of said component’s functionalities and structure.

### Reference

- [Reference](https://docs.google.com/document/d/1Bn9Bn0E9MZ1u6kLi7LYcQUdiczvGg9r-XEmOqetuclY/edit)

Assumptions
- A portion of necessitated componential APIs, functionalities, and structures likely lay undocumented/undiscovered
- Cross-component integration is not considered within the scope of this issue

### Acceptance

 - Transaction model created according to preparation document specification
 - Transaction CRUD API implemented according to preparation document specification 
 - Unit tests assessing Transaction CRUD API functionality, structure and instantiation created as covering at least 80% of viable code

## Tasks

- [x] Implement BPI Transaction model 
- [x] Implement BPI Transaction CRUD API
- [x] Implement tests for above functionalities

PR #533, opened 2022-10-01
Feature/532 bpi3 transactions model and crud api

ognjenkurtic:
# Description

Adds CUD API for transactions. Missing R actions and tests. WIll add those through subsequent PRs in order to reduce PR size.

## Related Issue

#532 

## Motivation and Context

Sets up the basis for working with transactions.

## How Has This Been Tested

Manual tests with Postman collection provided as part of the PR

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Request to be added as a Code Owner/Maintainer

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] I commit to abide by the Responsibilities of Code Owners/Maintainers.

PR #538, opened 2022-10-05
532 bri3 transactions crud api tests

ognjenkurtic:
# Description

Add unit tests for transactions CRUD Api. To be reviewed after PR #537 

## Related Issue

#532

## Motivation and Context

Tests the CRUD Api

## How Has This Been Tested

Unit tests

## Types of changes

## Checklist


Therecanbeonlyone1969 (commented):
Any comments would be related to comments raised in #533 and #537. So  this PR will change accordingly once requested changes are implemented.

ognjenkurtic:
@Therecanbeonlyone1969 All comments addressed. Please rereview.

examples/bri-3/src/bri/transactions/agents/mockTransactionStorage.agent.ts
```diff
@@ -0,0 +1,49 @@
+import { Injectable, NotFoundException } from '@nestjs/common';
```
Manik-Jain, line 1: since this is a mock agent, can we segregate this from main code and move inside a spec folder?

examples/bri-3/src/bri/transactions/agents/mockTransactionStorage.agent.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.transactionsStore.push(createdBp);
+
+    return Promise.resolve(createdBp);
+  }
+
+  async updateTransaction(transaction: Transaction): Promise<Transaction> {
+    const txToUpdate = this.transactionsStore.find(
+      (bp) => bp.id === transaction.id,
+    );
+    Object.assign(txToUpdate, transaction);
```
Manik-Jain, line 39: since the array is never updated, the getter calls from array will still return the previous value and not the updated one. Please update in the array also

examples/bri-3/src/bri/transactions/api/dtos/request/createTransaction.dto.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    const createTransactionDto = plainToInstance(CreateTransactionDto, dto);
+
+    // Act
+    const errors = await validate(createTransactionDto);
+
+    // Assert
+    expect(errors.length).toBe(1);
+    expect(errors[0].property).toEqual('id');
+    expect(errors[0].constraints.isNotEmpty).toContain(
+      'id should not be empty',
```
Manik-Jain, line 26: can we read this from constants file to avoid code change if this string is modified in actual code

examples/bri-3/src/bri/transactions/api/dtos/request/updateTransaction.dto.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    const updateTransactionDto = plainToInstance(UpdateTransactionDto, dto);
+
+    // Act
+    const errors = await validate(updateTransactionDto);
+
+    // Assert
+    expect(errors.length).toBe(1);
+    expect(errors[0].property).toEqual('payload');
+    expect(errors[0].constraints.isNotEmpty).toContain(
+      'payload should not be empty',
```
Manik-Jain, line 18: can we read this from constants file to avoid code change if this string is modified in actual code

examples/bri-3/src/bri/transactions/api/dtos/request/updateTransaction.dto.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    const updateTransactionDto = plainToInstance(UpdateTransactionDto, dto);
+
+    // Act
+    const errors = await validate(updateTransactionDto);
+
+    // Assert
+    expect(errors.length).toBe(1);
+    expect(errors[0].property).toEqual('signature');
+    expect(errors[0].constraints.isNotEmpty).toContain(
+      'signature should not be empty',
```
Manik-Jain, line 34: can we read this from constants file to avoid code change if this string is modified in actual code

examples/bri-3/src/bri/transactions/api/transactions.controller.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      const newTransactionId = await controller.createTransaction(requestDto);
+
+      // Act
+      const createdTransaction = await controller.getTransactionById(
+        newTransactionId,
+      );
+
+      // Assert
+      expect(createdTransaction.id).toEqual(requestDto.id);
```
Manik-Jain, line 73: can we not just assert that two objects are equal?

Manik-Jain (commented):
added some minor comments

examples/bri-3/src/bri/transactions/agents/mockTransactionStorage.agent.ts
ognjenkurtic: You usually want to keep the mocks next to the things they mock since that is the scope they are interested in. Similar logic to where we place the spec file.

examples/bri-3/src/bri/transactions/agents/mockTransactionStorage.agent.ts
ognjenkurtic: Then the update test would fail, wouldn't it? txToUpdate is storing a reference to the actual object, so updating it actually updates the underlying array.

examples/bri-3/src/bri/transactions/api/transactions.controller.spec.ts
ognjenkurtic: Could you please elaborate on what do you mean? They are not equal ATM because of the  TODOs on lines 81 and 82

examples/bri-3/src/bri/transactions/agents/mockTransactionStorage.agent.ts
Manik-Jain: yeah, but that doesn't stop us from structuring these inside a spec folder close to the class

examples/bri-3/src/bri/transactions/agents/mockTransactionStorage.agent.ts
Manik-Jain: yeah, that's why it has been failing for me

examples/bri-3/src/bri/transactions/api/dtos/request/updateTransaction.dto.spec.ts
ognjenkurtic: Since it is a slightly bigger change and the validation messages will change soon because we will revisit validaton as we are done with simple CRUD, i would leave it for the future. Is it ok if i open an issue for it for now?

examples/bri-3/src/bri/transactions/api/transactions.controller.spec.ts
Manik-Jain: const expected = {from : '', to : '', ...requestDto}

expect(expected).toEqual(createdTransaction)

examples/bri-3/src/bri/transactions/api/dtos/request/updateTransaction.dto.spec.ts
Manik-Jain: I'm fine a new issue..but I didn't understand why is it a big change to import the constant file and read from there?

ognjenkurtic: When the project is in it's early stages, i like to deffer decisions like these until there is a clearer picture of the exact needs. In other words, the dataset we have it to narrow to understand exactly what needs to be generalized, hence my proposal to open an issue and tackle it a later stage.

I introduced a constant file for this validation message. Please have a look.

examples/bri-3/src/bri/transactions/api/transactions.controller.spec.ts
ognjenkurtic: I would rather keep it line by line because mappping of 1 to 1 from the requestDto to createdTransaction will not be held for long. Example is status field which is not provided in the dto but exits in the final result. Also update transaction will never use this approach since the dto only contains a subset of fields. In my eyes, it is better to go with a simple line by line assertions for the exact fields we are trying to check instead of trying to introduce generalizations that are not complete - that will have many exceptions. Not against generalizations but they should come at a later stage.

Regarding the _expect(expected).toEqual(actual)_ it is in fact the other way around, _expect(actual).toEqual(expected)_

examples/bri-3/src/bri/transactions/api/dtos/request/updateTransaction.dto.spec.ts
ognjenkurtic: done

examples/bri-3/src/bri/transactions/api/dtos/request/createTransaction.dto.spec.ts
ognjenkurtic: done

examples/bri-3/src/bri/transactions/agents/mockTransactionStorage.agent.ts
ognjenkurtic: So this is an issue after the introduction of generic mapper and is solved on your branch, if i understand correctly

examples/bri-3/src/bri/transactions/agents/mockTransactionStorage.agent.ts
ognjenkurtic: storage mocks are a temporary solution as we learn prisma. They are cumbersome to maintain already and will become more so as we move forward. We ll probably have to ditch them and use a mocking framework instead. Also, we 've got spec files that do not have associated mock files, so having them in a separate folder sounds like an unnecessary step. I woould rather keep the state as it is atm, as we ll end up with a standardized approach of spec files next to the thing they test eventually. If you have a strong motivation for this change i can do it, but i do not see additional value in doing it.

examples/bri-3/src/bri/shared/constants.ts
```diff
@@ -0,0 +1,2 @@
+// DTO validation messages
```
Manik-Jain, line 1: not a blocker, just a suggestion -> how about using `common` instead of `shared`

examples/bri-3/src/bri/transactions/api/transactions.controller.spec.ts
Manik-Jain: ok

examples/bri-3/src/bri/shared/constants.ts
ognjenkurtic: For me it is completely the same. Not sure what others think.

Manik-Jain (dismissed):

skosito (dismissed):
just 2 double checks:
- do we need dev.db?
- do we need migration for transactions table?

approved

ognjenkurtic:
@skosito @Manik-Jain please reapprove, just addressed comments from skosito.

Manik-Jain (approved):

Manik-Jain:
> @skosito @Manik-Jain please reapprove, just addressed comments from skosito.

Done

skosito (approved):

Kasshern (approved):

Issue #540, opened 2022-10-06
SRI Subject Account Component- Interfaces, Classes and API


skosito:
## Overview

Integrate with BPI Account and BPI Subject

### Reference

Check issue https://github.com/eea-oasis/baseline/issues/489 for more context

## Tasks

- [x] Implement BPI Subject Account model through interface and class that implements it
- [x] Implement BPI Subject Account CRUD
- [x] Implement BPI Subject Account tests

PR #541, opened 2022-10-06
Feature/489 bpi account crud api

skosito:
# Description

NOTE: I will add many-to-many relationship with bpi subject account in follow up PR, this is simple set up of bpi accounts CRUD as first step - opened new issue for it https://github.com/eea-oasis/baseline/issues/540

## Related Issue

https://github.com/eea-oasis/baseline/issues/489

## Types of changes

## Checklist


examples/bri-3/src/bri/identity/bpiAccounts/agents/bpiAccounts.agent.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { NOT_FOUND_ERR_MESSAGE } from '../api/err.messages';
+import { BpiAccount } from '../models/bpiAccount';
+
+import { BpiAccountStorageAgent } from './bpiAccountsStorage.agent';
+
+// Agent methods have extremely declarative names and perform a single task
+@Injectable()
+export class BpiAccountAgent {
+  constructor(private repo: BpiAccountStorageAgent) {}
```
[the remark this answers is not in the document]
Manik-Jain: @ognjenkurtic do we have a code recommendation document?

examples/bri-3/src/bri/identity/bpiAccounts/agents/bpiAccounts.agent.ts
```diff
@@ -0,0 +1,48 @@
...
+  public throwIfCreateBpiAccountInputInvalid() {
+    return true;
+  }
+
+  public createNewBpiAccount(nonce: string): BpiAccount {
+    return new BpiAccount(v4(), nonce, []);
+  }
+
+  public async fetchUpdateCandidateAndThrowIfUpdateValidationFails(
```
Manik-Jain, line 21: I'm actually not sure if this needs to explicitely specified in the method name to `throwIfUpdateValidationFails`. I mean isn't it implicit that the code should throw exception if it doesn't find something?

examples/bri-3/src/bri/identity/bpiAccounts/agents/bpiAccounts.agent.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    if (!bpiAccountToUpdate) {
+      throw new NotFoundException(NOT_FOUND_ERR_MESSAGE);
+    }
+
+    return bpiAccountToUpdate;
+  }
+
+  public updateBpiAccount(bpiAccountToUpdate: BpiAccount, nonce: string) {
+    bpiAccountToUpdate.updateNonce(nonce);
```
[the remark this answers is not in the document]
Manik-Jain: +1 Same question. If there is no other use case, do we even need the nonce as a parameter?

examples/bri-3/src/bri/identity/bpiAccounts/agents/bpiAccounts.agent.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  public async fetchDeleteCandidateAndThrowIfDeleteValidationFails(
```
Manik-Jain, line 37: I'm actually not sure if this needs to explicitely specified in the method name to throwIfUpdateValidationFails. I mean isn't it implicit that the code should throw exception if it doesn't find something?

examples/bri-3/src/bri/identity/bpiAccounts/agents/bpiAccountsStorage.agent.ts
```diff
@@ -0,0 +1,56 @@
+import { Injectable, NotFoundException } from '@nestjs/common';
+import { PrismaService } from '../../../../../prisma/prisma.service';
+import { NOT_FOUND_ERR_MESSAGE } from '../api/err.messages';
+import { BpiAccount } from '../models/bpiAccount';
+
+// Repositories are the only places that talk the Prisma language of models.
```
Manik-Jain, line 6: @ognjenkurtic Quick question, If this is a prisma repository, why are we naming it as `StorageAgent`, and not just `BpiAccountRepo`

examples/bri-3/src/bri/identity/bpiAccounts/agents/bpiAccounts.agent.ts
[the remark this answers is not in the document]
ognjenkurtic: not yet, there is an open issue to add a readme file but we didnt get to it yet.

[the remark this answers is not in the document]
ognjenkurtic: There are docs for branch naming and commit messages through, please look into these first.

examples/bri-3/src/bri/identity/bpiAccounts/agents/bpiAccounts.agent.ts
ognjenkurtic: no, it could return null and let the caller deal with the situation. Declarative naming of the agent methods helps the reader of the commandhandler to understand the execution flow even without looking into the agent  implementations.

examples/bri-3/prisma/schema.prisma
```diff
@@ -18,6 +18,11 @@ model BpiSubject {
   publicKey   String
 }
 
+model BpiAccount {
+  id    String @id @default(uuid())
+  nonce String
```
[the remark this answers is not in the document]
Therecanbeonlyone1969: @ognjenkurtic @skosito @Manik-Jain @Kasshern @Ybittan ... we need to agree on the nonce typing. `int` is easiest,  or `string` is ok as well to give more flexibility. for BRI-3 I would recommend sticking with `int` for simplicity's sake ... this would allow for example synchronization with e.g. an Ethereum account if desired.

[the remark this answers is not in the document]
Manik-Jain: I agree with `int` too.

Also, if we want to be future safe and handle any large numbers, we can also consider incorporating number or bigInt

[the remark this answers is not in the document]
skosito: it was int initially but comment was to move to string here https://github.com/eea-oasis/baseline/pull/507#discussion_r956461635
i will revert to int

examples/bri-3/src/bri/identity/bpiAccounts/capabilities/createBpiAccount/createBpiAccountCommand.handler.ts
```diff
@@ -0,0 +1,28 @@
...
+import { BpiAccountStorageAgent } from '../../agents/bpiAccountsStorage.agent';
+import { CreateBpiAccountCommand } from './createBpiAccount.command';
+
+@CommandHandler(CreateBpiAccountCommand)
+export class CreateBpiAccountCommandHandler
+  implements ICommandHandler<CreateBpiAccountCommand>
+{
+  constructor(
+    private agent: BpiAccountAgent,
+    private repo: BpiAccountStorageAgent,
```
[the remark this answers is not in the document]
ognjenkurtic: repo => storageAgent

examples/bri-3/test/bri.postman_collection.json
```diff
@@ -446,129 +446,6 @@
 					]
 				}
 			]
-		},
-		{
-			"name": "Transaction",
```
ognjenkurtic, line 329: why is transaction removed from the postman collection?

ognjenkurtic (approved):

examples/bri-3/src/bri/identity/bpiAccounts/agents/bpiAccountsStorage.agent.ts
Kasshern: I believe we originally used repo and decided to move over to agent to correlate functionality between the two injectable agents.  I also remember a discussion of the word "repo" possibly introducing confusion upon first look at the repository code.

examples/bri-3/src/bri/identity/bpiSubjects/subjects.module.ts
```diff
@@ -8,25 +8,40 @@ import { GetAllBpiSubjectsQueryHandler } from './capabilities/getAllBpiSubjects/
...
+import { BpiAccountAgent } from '../bpiAccounts/agents/bpiAccounts.agent';
+import { BpiAccountStorageAgent } from '../bpiAccounts/agents/bpiAccountsStorage.agent';
 
 export const CommandHandlers = [
   CreateBpiSubjectCommandHandler,
   UpdateBpiSubjectCommandHandler,
   DeleteBpiSubjectCommandHandler,
+  CreateBpiAccountCommandHandler,
+  UpdateBpiAccountCommandHandler,
+  DeleteBpiAccountCommandHandler,
```
Kasshern, line 26: @skosito What is the reasoning behind listing all of these handles and agents of the BpiAccount here in the subjects module rather than the accounts module?

Kasshern (dismissed):
@skosito Left a question or two for whenever you have time. Getting a warning about prisma version update required. Should resolve if main is pulled in here.

examples/bri-3/src/bri/identity/bpiAccounts/capabilities/getAllBpiAccounts/getAllBpiAccountQuery.handler.ts
```diff
@@ -0,0 +1,23 @@
...
+export class GetAllBpiAccountsQueryHandler
+  implements IQueryHandler<GetAllBpiAccountsQuery>
+{
+  constructor(private readonly storageAgent: BpiAccountStorageAgent) {}
+
+  async execute() {
+    const bpiSubjects = await this.storageAgent.getAllBpiAccounts();
+
+    return bpiSubjects.map((bp) => {
+      return {
```
Manik-Jain, line 16: We can use current mapper here

Manik-Jain (commented):
Just a small change request

examples/bri-3/src/bri/identity/bpiSubjects/subjects.module.ts
skosito: good point, i think i fixed that in other PR

examples/bri-3/src/bri/identity/bpiAccounts/capabilities/getAllBpiAccounts/getAllBpiAccountQuery.handler.ts
skosito: @Manik-Jain i think it would be better to leave todo until we decide which mapper to use? if we use automapper we will have to change this too

ognjenkurtic: Yeah, let's not duplicate work if the new mapper is coming.

examples/bri-3/prisma/schema.prisma
```diff
@@ -18,6 +18,11 @@ model BpiSubject {
   publicKey   String
 }
 
+model BpiAccount {
```
Therecanbeonlyone1969, line 21: @skosito 
- The model structure is defined in [R251](https://github.com/eea-oasis/baseline-standard/blob/main/core/baseline-core-v1.0-psd01.md#r251) ... the model (and migration above) does not comply with that. The value account balance(s) is actually incorrect because of D35.  
- Also, if we do not define the relationship between `BPIAccount` and its state here, how do we intend to achieve that? IMHO, just put a pointer object to the state object here e.g. could be a smart contract on Ethereum or on the BPIs zkEVM
- Also, we do not have a place for the required vector commitment scheme for the BPI Account state? In other words, where do we intend to store it how do we construct it? IMHO, the easiest way is to represent it is through a merkle tree where each BPI Account is a leaf, that kills all bords with one so to speak

Based on your suggestions the handlers etc below change too

cc @ognjenkurtic @Kasshern @Manik-Jain @Ybittan

Therecanbeonlyone1969 (changes requested):

examples/bri-3/prisma/schema.prisma
skosito: @Therecanbeonlyone1969 this is based on already approved model  https://github.com/eea-oasis/baseline/pull/507 and discussion here https://docs.google.com/document/d/1DCi0Ap9EK2cBxxEVzmOPkdkcl1hnv7-qoELGHODb1PM/edit where the model for first version is clearly defined and approved

if we follow the process i suggest we merge current work and open issue for additional changes - there are 3 PRs waiting to be merged related to accounts and bpi subject accounts with relations between them defined which are correct, if i understand your comments i need to add more fields to model and db, which can be follow up work and will result in additional migrations

skosito: on second look, a lot of these fields to add would be placeholders anyways, because we dont have adequate db models to connect them, so we can just add them as we go

ognjenkurtic: @Therecanbeonlyone1969 yes, let's please stick with the proposed models from the document @skosito outlined. As soon as CRUD milestone is reached we will start tackling all the intricate details of state storage and other MUST requirements outlined in the standard. These will be tackled in separate smaller issues to avoid tackling mulitple complex topics at the same time and to enable detailed discussions in the issues before implementation even begins. Dont worry,. no MUST requirement outlined in the scope document  will be left behind ;).

Therecanbeonlyone1969: Ok. That is fine. It just needs to be tracked

Therecanbeonlyone1969 (dismissed):

ognjenkurtic (dismissed):

Kasshern (approved):

Therecanbeonlyone1969 (approved):
Reapprove after path change

Issue #552, opened 2022-10-14
[SRI] Refactor transaction API with mapper


Manik-Jain:
## Overview

Currently Transaction API doesn't use mapper strategy. This task aims to refactor transaction API with mapper

### Reference
https://github.com/eea-oasis/baseline/pull/528

### Acceptance

Transaction API shall be using mappers

## Tasks

- [ ] A list of tasks that need
- [ ] to be done to call this issue done

Manik-Jain:
PR : https://github.com/eea-oasis/baseline/pull/565 raised for this change

Manik-Jain:
PR https://github.com/eea-oasis/baseline/pull/565 merged

Issue #553, opened 2022-10-14
[SRI] modify mapper to invoke class constructor dynamically


Manik-Jain:
## Overview

Mapper code needs a little modification to dynamically invoke class constructors

### Assumptions

All constructor parameters are well defined in classes

### Acceptance

class constructors are invoked dynamically instead of explicit invocation 

## Tasks

- [ ] A list of tasks that need
- [ ] to be done to call this issue done

Manik-Jain:
not needed anymore as we have migrated to Automapper

PR #557, opened 2022-10-17
adding changes to refactor transaction api

Manik-Jain:
# Description

Changes made to refactor Transaction API to use generic mapper

## Related Issue

https://github.com/eea-oasis/baseline/issues/552

## Motivation and Context

In order to have code sync across modules, we need to use generic mapper for all APIs

## How Has This Been Tested

tested locally, and via test cases

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/examples/bri-3/src/bri/transactions/agents/transactionStorage.agent.ts
+++ b/examples/bri-3/src/bri/transactions/agents/transactionStorage.agent.ts
@@ -1,30 +1,43 @@
 import { Injectable, NotFoundException } from '@nestjs/common';
+import Mapper from '../../utils/mapper';
 import { PrismaService } from '../../../../prisma/prisma.service';
 import { NOT_FOUND_ERR_MESSAGE } from '..//api/err.messages';
 import { Transaction } from '../models/transaction';
+import { getType } from 'tst-reflect';
 
 @Injectable()
 export class TransactionStorageAgent extends PrismaService {
+  constructor(private readonly mapper: Mapper) {
+    super();
+  }
+
+  async getAllTransactions(): Promise<Transaction[]> {
+    const transactionModels = await this.transaction.findMany();
+    return transactionModels.map((transactionModel) => {
+      return this.mapper.map(transactionModel, getType<Transaction>(), {
+        opts: {
+          from: null, // TODO: transactionModel.from once BpiAccount in the prisma schema,
+          to: null, // TODO: transactionModel.from once BpiAccount in the prisma schema,
+        },
+      }) as Transaction;
+    });
+  }
+
   async getTransactionById(id: string): Promise<Transaction> {
     const transactionModel = await this.transaction.findUnique({
-      where: { id: id },
+      where: { id },
     });
 
     if (!transactionModel) {
       throw new NotFoundException(NOT_FOUND_ERR_MESSAGE);
     }
 
-    return new Transaction(
-      transactionModel.id,
-      transactionModel.nonce,
-      transactionModel.workflowInstanceId,
-      transactionModel.workstepInstanceId,
-      null, // TODO: transactionModel.from once BpiAccount in the prisma schema,
-      null, // TODO: transactionModel.to once BpiAccount in the prisma schema,
-      transactionModel.payload,
-      transactionModel.signature,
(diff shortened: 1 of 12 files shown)


skosito:
@Manik-Jain what about automapper, is there a decision? imo we should decide on that before adding additional code on old mapper

examples/bri-3/src/bri/transactions/agents/transactions.agent.ts
```diff
@@ -58,7 +58,17 @@ export class TransactionAgent {
       throw new BadRequestException(UPDATE_WRONG_STATUS_ERR_MESSAGE);
     }
 
-    return transactionToUpdate;
+    return new Transaction(
```
ognjenkurtic, line 61: why is this needed when there is already a mapping from prisma model to domain object in the storage agent?

examples/bri-3/src/bri/transactions/agents/transactionStorage.agent.ts
```diff
@@ -1,30 +1,43 @@
...
 import { Transaction } from '../models/transaction';
+import { getType } from 'tst-reflect';
 
 @Injectable()
 export class TransactionStorageAgent extends PrismaService {
+  constructor(private readonly mapper: Mapper) {
+    super();
+  }
+
+  async getAllTransactions(): Promise<Transaction[]> {
```
ognjenkurtic, line 14: Just a note, we deliberately left our the get all endpoint because we are unsure of the scope for this query. I am fine with including it here.

examples/bri-3/src/bri/transactions/capabilities/updateTransaction/updateTransactionCommand.handler.ts
```diff
@@ -10,6 +12,7 @@ export class UpdateTransactionCommandHandler
   constructor(
     private agent: TransactionAgent,
     private storageAgent: TransactionStorageAgent,
+    private readonly mapper: Mapper,
```
ognjenkurtic, line 15: mapper is not used anywhere

examples/bri-3/src/bri/transactions/agents/transactions.agent.ts
Manik-Jain: this is needed till the time https://github.com/eea-oasis/baseline/issues/553 is resolved, or we have auto-mapper in place, whatever comes first. This will be changed once we have that in place

examples/bri-3/src/bri/transactions/capabilities/updateTransaction/updateTransactionCommand.handler.ts
Manik-Jain: change done

examples/bri-3/src/bri/transactions/agents/transactionStorage.agent.ts
Manik-Jain: > @Manik-Jain what about automapper, is there a decision? imo we should decide on that before adding additional code to current mapper

Yes, we have agreed to change response interfaces to classes and then we can have auto-mapper changes. 
https://github.com/eea-oasis/baseline/issues/558

Manik-Jain:
> @Manik-Jain what about automapper, is there a decision? imo we should decide on that before adding additional code to current mapper

Yes, we have agreed to change response interfaces to classes and then we can have auto-mapper changes.
https://github.com/eea-oasis/baseline/issues/558

ognjenkurtic (approved):

Therecanbeonlyone1969:
@Manik-Jain @skosito @ognjenkurtic should we then wait until we have the automapper properly implemented before we do this PR?

cc @Kasshern @Ybittan

ognjenkurtic:
> @Manik-Jain @skosito @ognjenkurtic should we then wait until we have the automapper properly implemented before we do this PR?
> 
> cc @Kasshern @Ybittan

@Manik-Jain is best to provide the answer here. If we can do automapper quickly i guess we do not need to work on the old mapper.

Manik-Jain:
Change taken care in automapper PR

PR #562, opened 2022-10-20
Bpi account and subject account tests

skosito:
# Description

Unit tests for api from issues https://github.com/eea-oasis/baseline/issues/540 and https://github.com/eea-oasis/baseline/issues/489
After this is merged both issues can be closed.

## Related Issue

https://github.com/eea-oasis/baseline/issues/489 and https://github.com/eea-oasis/baseline/issues/540

## Motivation and Context

Unit tests testing not just entities in isolation but also their relation.

## How Has This Been Tested

npm run test passing

## Types of changes

## Checklist


examples/bri-3/src/bri/identity/bpiAccounts/api/accounts.controller.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+import { SubjectController } from '../../bpiSubjects/api/subjects.controller';
+import { SubjectAccountController } from '../../bpiSubjectAccounts/api/subjectAccounts.controller';
+import { CreateBpiSubjectDto } from '../../bpiSubjects/api/dtos/request/createBpiSubject.dto';
+import { CreateBpiSubjectAccountDto } from '../../bpiSubjectAccounts/api/dtos/request/createBpiSubjectAccount.dto';
+import { SubjectAccountModule } from '../../bpiSubjectAccounts/subjectAccounts.module';
+import { validate as uuidValidate, version as uuidVersion } from 'uuid';
+import { BpiSubjectStorageAgent } from '../../bpiSubjects/agents/bpiSubjectsStorage.agent';
+import { MockBpiSubjectStorageAgent } from '../../bpiSubjects/agents/mockBpiSubjectStorage.agent';
+
+describe.only('AccountController', () => {
```
ognjenkurtic, line 31: .only should be removed

examples/bri-3/src/bri/identity/bpiAccounts/api/accounts.controller.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+  let accountController: AccountController;
```
ognjenkurtic, line 32: I used the approach where i registered only the mock storage from other modules and creating the necessary data directly in it, to avoid going through the separate controller as it is already tested in it's own test file. You can see an example in messages spec file. Not sure which approach is better but we should definitely stick to one of them. I am lightly inclined to my approach as the unit tests are then limited only to the scope of the single controller.

ognjenkurtic: and you need less stuff to initialize the test module

skosito: ok i'll change

ognjenkurtic (approved):

Issue #590, opened 2022-11-11
[SRI] login with Metamask minimalistic UI


Manik-Jain:
## Overview

In order for BpiSubjects to authenticate themselves, we need to expose a basic ReactJs style UI that allows subjects to sign a MM message. The signed message will be sent to the backend API and a jwt token will be assigned to the authenticated users

### Acceptance

Upon a successful authentication, a jwt token shall be assigned to the authenticated users

Issue #589, opened 2022-11-11
[SRI] expose authentication API


Manik-Jain:
## Overview

In order for users to authenticate themselves, we need to expose public endpoints that allow them to login, and request for jwt token. This will be invoked from UI upon a bip39 string is signed and signature is sent to the backend

### Acceptance

Authenticated BpiSubjects should be able to get a jwt token

Issue #588, opened 2022-11-11
[SRI] implement NestJS style authentication


Manik-Jain:
## Overview

As discussed with the SRI team, we need to implement NestJs style authentication. Based on the BpiSubject public key, we will assign a JWT token to the authenticated subjects.

### Reference

https://docs.google.com/document/d/1A4wr7Vqt9mUoDezKI5njk6eKRixGP1nUgxUe_E7pay0/edit

### Acceptance

None of the endpoints shall be accessible without a jwt token

## Tasks

1. Jwt passport
2. Jwt strategy
3. Jwt guards

PR #592, opened 2022-11-11
Sri auth

Manik-Jain:
# Description

This PR has changes for SRI authentication - backend and ReactJs frontend

## Related Issue

https://github.com/eea-oasis/baseline/issues/588
https://github.com/eea-oasis/baseline/issues/589
https://github.com/eea-oasis/baseline/issues/590

## Motivation and Context

In order to only authenticated access, we need to have a mandatory bearer token for each end-point

## How Has This Been Tested

Local tested

## Types of changes

## Checklist

-[X] I commit to abide by the Responsibilities of Code Owners/Maintainers.


examples/bri-3/src/bri/auth/api/auth.controller.ts
```diff
@@ -1,14 +1,21 @@
...
   @Public()
   @Post('login')
   async login(@Request() req) {
     return this.authService.login(req.body);
   }
+
+  @Public()
+  @Get('mnemonic')
+  async mnemonic(@Request() req) {
+    return generateMnemonic();
```
ognjenkurtic, line 19: If someone knows a message and a signature of that message for a specific public key, they can pass the login as that public key. This is because the mnemonic (one time pass)  is not stored during geneation so we do not check if it was issued by our backend during login. Do we really want this?

@Manik-Jain @skosito @Therecanbeonlyone1969 @Kasshern @biscuitdey @Ybittan

examples/bri-3/front-end/src/login/Login.js
```diff
@@ -0,0 +1,35 @@
...
+      message: mnemonic.data,
+      signature,
+      publicKey: accounts.result[0],
+    };
+
+    const tokenRequest = await axios.post(`${SRI_BACKEND}/auth/login`, input);
+
+    if (tokenRequest?.data?.access_token) {
+      console.log(tokenRequest.data.access_token);
+      localStorage.setItem('token', tokenRequest.data.access_token);
```
skosito, line 24: in general its bad practice to store token in local storage but i guess it will do for this

examples/bri-3/package.json
```diff
@@ -31,11 +31,19 @@
     "@nestjs/common": "^9.0.0",
     "@nestjs/core": "^9.0.0",
     "@nestjs/cqrs": "^9.0.1",
+    "@nestjs/jwt": "^9.0.0",
```
skosito, line 34: still reviewing but 10 more dependencies for this seems a lot, why we need passport and nest jwt?

examples/bri-3/src/bri/auth/agents/authAccessor.agent.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from '@nestjs/common';
+import { PrismaService } from '../../../../prisma/prisma.service';
+
+@Injectable()
+export class AuthAccesorService extends PrismaService {
```
skosito, line 5: i dont think new service is needed, this looks like it should be part of bpi subject storage agent
authAccessor.findOne doesnt say anything about bpi subject being returned

examples/bri-3/src/bri/auth/api/auth.controller.ts
skosito: yea... we need to store some nonce that is signed and then when authenticating pull the last nonce from db, not send it in request body
also, why using bip39 mnemonic at first place as thing to be signed? lets just store some incrementing nonce or uuid or something like that...

examples/bri-3/src/bri/auth/guards/sri-jwt.guard.ts
```diff
@@ -0,0 +1,22 @@
...
+import { IS_PUBLIC_KEY } from 'src/bri/decorators/public-endpoint';
+
+@Injectable()
+export class JwtAuthGuard extends AuthGuard('jwt') {
+  constructor(private reflector: Reflector) {
+    super();
+  }
+
+  canActivate(context: ExecutionContext) {
+    const isPublic = this.reflector.getAllAndOverride<boolean>(IS_PUBLIC_KEY, [
```
skosito, line 13: what is this doing? please add comment

examples/bri-3/src/bri/auth/guards/sri-local.guard.ts
```diff
@@ -0,0 +1,5 @@
+import { Injectable } from '@nestjs/common';
+import { AuthGuard } from '@nestjs/passport';
+
+@Injectable()
+export class SriLocalAuthGuard extends AuthGuard('local') {}
```
skosito, line 5: why we need both jwt and local auth guards?

examples/bri-3/src/bri/auth/services/auth.service.ts
```diff
@@ -0,0 +1,35 @@
...
+@Injectable()
+export class AuthService {
+  constructor(
+    private readonly authAccesorService: AuthAccesorService,
+    private jwtService: JwtService,
+    private readonly loginService: LoginService,
+  ) {}
+
+  async validateUser(username: string, pass: string): Promise<any> {
+    const user = await this.authAccesorService.findOne(username);
```
skosito, line 17: use bpiSubject terminology instead of user, use publicKey instead of username

why do we need password at all? or verifyUser method?

examples/bri-3/src/bri/auth/model/login.ts
```diff
@@ -0,0 +1,8 @@
+export interface Login {
+  //bip39 string signed from Metamask
+  message: string;
```
skosito, line 3: what is Login, is it dto or? if its request body, please use dto convention we use, with not empty validators etc
also message should not be part of this request body, but pulled out of db, check comments above

examples/bri-3/src/bri/auth/services/auth.service.ts
```diff
@@ -0,0 +1,35 @@
...
+import { JwtService } from '@nestjs/jwt';
+import { AuthAccesorService } from '../agents/authAccessor.agent';
+import { errorMessage } from '../constants';
+import { Login } from '../model/login';
+import { LoginService } from './login.service';
+
+@Injectable()
+export class AuthService {
+  constructor(
+    private readonly authAccesorService: AuthAccesorService,
```
skosito, line 11: we introduce 4 services that are at least somehow  connected
- auth accessor service
- jwt service
- auth service
- login service

why so many services for simple auth flow?

examples/bri-3/src/bri/auth/services/login.service.ts
```diff
@@ -0,0 +1,9 @@
+import { Injectable } from '@nestjs/common';
+import { ethers } from 'ethers';
+
+@Injectable()
+export class LoginService {
```
skosito, line 5: see my comment about 4 services for auth above...

examples/bri-3/src/bri/auth/strategy/jwt.strategy.ts
```diff
@@ -0,0 +1,20 @@
+import { ExtractJwt, Strategy } from 'passport-jwt';
```
skosito, line 1: importing nestjs/passport
passport
passport-jwt
nestjs/jwt

in same project? do we need all those packages?

examples/bri-3/src/bri/auth/strategy/local.strategy.ts
```diff
@@ -0,0 +1,16 @@
+import { Strategy } from 'passport-local';
+import { PassportStrategy } from '@nestjs/passport';
+import { Injectable, UnauthorizedException } from '@nestjs/common';
+import { AuthService } from '../services/auth.service';
+
+//use this to check if the user exists in DB, and assign a jwt token if it exists
+@Injectable()
+export class LocalStrategy extends PassportStrategy(Strategy) {
```
skosito, line 8: what is the difference between local and jwt strategies? why do we need both?

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjectsStorage.agent.ts
```diff
@@ -27,6 +27,7 @@ export class BpiSubjectStorageAgent extends PrismaService {
 
   async getAllBpiSubjects(): Promise<BpiSubject[]> {
     const bpiSubjectModels = await this.bpiSubject.findMany();
+    console.log('bpiSubjectModels', bpiSubjectModels);
```
skosito, line 30: please remove console.logs before merging

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjectsStorage.agent.ts
```diff
@@ -44,6 +45,7 @@ export class BpiSubjectStorageAgent extends PrismaService {
   }
 
   async createNewBpiSubject(bpiSubject: BpiSubject): Promise<BpiSubject> {
+    bpiSubject.publicKey = bpiSubject.publicKey.toLowerCase();
```
skosito, line 48: no need to store as lower case, only lower case both on places we compare

examples/bri-3/src/bri/decorators/public-endpoint.ts
```diff
@@ -0,0 +1,4 @@
+import { SetMetadata } from '@nestjs/common';
```
skosito, line 1: what is public endpoint?

skosito (changes requested):

examples/bri-3/src/bri/decorators/public-endpoint.ts
Manik-Jain: in order for login, or any, function to be invoked without any bearer token we need to have a way. This decorator does that job for us.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjectsStorage.agent.ts
Manik-Jain: as a general observation, ethers and other library returns us the lowercase MM public key.

examples/bri-3/src/bri/auth/strategy/local.strategy.ts
Manik-Jain: js doc was already added to the classes.

examples/bri-3/src/bri/auth/strategy/jwt.strategy.ts
Manik-Jain: I would suggest reading through this document: https://docs.nestjs.com/security/authentication#authentication

examples/bri-3/src/bri/auth/services/auth.service.ts
Manik-Jain: No, they are not at all connected. Each service has its own defined responsibility.

examples/bri-3/package.json
Manik-Jain: kindly skim read here: https://docs.nestjs.com/security/authentication#authentication

examples/bri-3/src/bri/auth/agents/authAccessor.agent.ts
Manik-Jain: I wanted to keep auth related code inside `auth` folder, and leave other CRUD operations inside identity. I don't this change is causing any blockers

examples/bri-3/src/bri/auth/guards/sri-local.guard.ts
Manik-Jain: Right, we just need jwt guard. Removing the other one

examples/bri-3/src/bri/auth/services/auth.service.ts
Manik-Jain: Removed

examples/bri-3/src/bri/auth/model/login.ts
Manik-Jain: changed

examples/bri-3/front-end/src/login/Login.js
Manik-Jain: Well, there is a mixed approach that people/teams use. Anyways, for the current scope, this can be removed

examples/bri-3/src/bri/auth/api/auth.controller.ts
Manik-Jain: @ognjenkurtic Right, I missed on the OTP storage part. Will push in the code soon.
@skosito the incrementing nonce approach works, and I think this is legacy. bip39 mnemonics can act as the random string of words to be signed, and updated in DB  the hash for each new generated mnemonic, similar to a nonce.

Another approach to it can be, we store in the signature instead of the nonce/mnemonic, to identify if a jwt token was linked to the provided signature. This way, we cover both the scenario

ognjenkurtic (changes requested):
Some general comments:

1) Please follow the branch naming standard of the Baseline repository.

2) Please provide more meaningful commit messages. General agreement is to do "This commit does..." approach where your commit message just continues the sentence. Also, commits should be much smaller.

3) Please follow the process we agreed on in the beginning of the project for defining the implementation direction. 

           Create a doc with proposal -> Iterate with the team -> Write down final agreement -> Create separate smaller issues for implementation -> Implement by doing a single PR per issue

This is important for a multitude of reasons, most notably:
   - We document our design decisions in a transparent and consistent way
   - We do most of the design during the initial iterations by the team, avoiding costly rewrites and discussions once PR is created
   - We produce smaller PRs, easier to review thus with higher quality

4) Please follow the agreed architecture of the project - I believe there is no need to explain why.

As we are limited with resources and are trying to fit a lot of work til the end of the year, following the process is essential for keeping things on time and producing high quality output.

examples/bri-3/src/bri/auth/api/auth.controller.ts
Therecanbeonlyone1969: I agree a uuid is sufficient as a secret to be signed. The secret should be encrypted to the public key of the requesting user such that decryption requires the corresponding private key.

For replay protection, we need an incrementing nonce and the specific BPI domain to be accessed to be returned in the query response with the JWT, ideally as part of the JWT.

examples/bri-3/src/bri/decorators/public-endpoint.ts
Therecanbeonlyone1969: @Manik-Jain so empty meta data signifies a login attempt?

examples/bri-3/src/bri/auth/api/auth.controller.ts
```diff
@@ -0,0 +1,21 @@
...
+import { generateMnemonic } from 'bip39';
+
+@Controller('auth')
+export class AuthController {
+  constructor(private readonly authService: AuthService) {}
+
+  @Public()
+  @Post('login')
+  async login(@Request() req) {
+    return this.authService.login(req.body);
```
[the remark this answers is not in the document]
Therecanbeonlyone1969: agree ... please, stick to the agreed-upon pattern @Manik-Jain, we need to be 100% constant throughout the implementation otherwise we are building in maintenance nightmares when it comes to trouble shooting, upgrades etc.

examples/bri-3/package.json
```diff
@@ -31,11 +31,19 @@
...
     "@prisma/client": "^4.4.0",
+    "bip39": "^3.0.4",
     "class-transformer": "^0.5.1",
     "class-validator": "^0.13.2",
+    "eslint-config-react-app": "^7.0.1",
+    "ethers": "^5.7.2",
     "ew-precise-proofs-js": "^1.2.0",
+    "passport": "^0.6.0",
+    "passport-jwt": "^4.0.0",
+    "passport-local": "^1.0.0",
```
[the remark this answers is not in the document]
Therecanbeonlyone1969: @ognjenkurtic unless I read the docs wrong, you need local to extend the chosen strategy, namely jwt, in order to be able to do relevant configurations of the strategy through injecting a metadata object in the strategy -- which we will probably want to do soon since we need resolvable user ids. Hence, I think it is advisable to just do the extension pattern at this point. thoughts? cc @skosito @Manik-Jain @Kasshern

examples/bri-3/src/bri/auth/strategy/jwt.strategy.ts
```diff
@@ -0,0 +1,20 @@
...
+  constructor() {
+    super({
+      jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+      ignoreExpiration: false,
+      secretOrKey: jwtConstants.secret,
+    });
+  }
+
+  async validate(payload: any) {
+    return { userId: payload.sub, username: payload.username };
```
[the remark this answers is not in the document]
Therecanbeonlyone1969: @ognjenkurtic if the id is resolvable to its controlling keys then id is enough. Otherwise, you need to indicate which public key controls the id. I prefer the former over the latter approach.

examples/bri-3/src/main.ts
```diff
@@ -5,6 +5,7 @@ import { AppModule } from './app.module';
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
   app.useGlobalPipes(new ValidationPipe());
+  app.enableCors();
```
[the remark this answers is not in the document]
Therecanbeonlyone1969: yes, super important ... please, open tracking issues to any open To-Do items. @Manik-Jain

examples/bri-3/src/bri/decorators/public-endpoint.ts
Manik-Jain: @Therecanbeonlyone1969 no, any end-points annotated with this decorator will not be picked by the Nest authenticator. 

From proposal doc, this decorator covers the proposed approach
`NestJs provides a capability to either manually update each end-point to run in an authenticated manner, or we can make all the end-points authenticated by default and choose which ones can be allowed public access. I propose to use the second approach`

examples/bri-3/src/bri/auth/api/auth.controller.ts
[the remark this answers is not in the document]
Manik-Jain: Right, this was a miss on my end. This will be pushed in the next commit

examples/bri-3/front-end/README.md
```diff
@@ -0,0 +1,70 @@
+# Getting Started with Create React App
```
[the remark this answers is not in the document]
Manik-Jain: @ognjenkurtic The agreement on the call was to use a minimalistic UI to accomplish the login with MM capability. I'm not understanding what is missed here if we have a simple React application that can be extended further down the line. Could you elaborate what is the issue here, or why are we talking about Vanila JS here?

examples/bri-3/package.json
```diff
@@ -31,11 +31,19 @@
...
     "@nestjs/core": "^9.0.0",
     "@nestjs/cqrs": "^9.0.1",
+    "@nestjs/jwt": "^9.0.0",
+    "@nestjs/passport": "^9.0.0",
     "@nestjs/platform-express": "^9.0.0",
     "@prisma/client": "^4.4.0",
+    "bip39": "^3.0.4",
     "class-transformer": "^0.5.1",
     "class-validator": "^0.13.2",
+    "eslint-config-react-app": "^7.0.1",
```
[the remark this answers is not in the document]
Manik-Jain: please refer to the commit here: https://github.com/eea-oasis/baseline/pull/592/commits/2367099fe113baadbdcc561f0f5359156fb74f4f

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjectsStorage.agent.ts
```diff
@@ -64,4 +65,15 @@ export class BpiSubjectStorageAgent extends PrismaService {
       where: { id: bpiSubject.id },
     });
   }
+
+  async getByPublicKey(publicKey: string): Promise<BpiSubject> {
```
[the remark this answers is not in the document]
Manik-Jain: Do you mean BpiSubjectAgent can return anything else besides a BpiSubject? If yes, I agree to the suggested name.

examples/bri-3/src/main.ts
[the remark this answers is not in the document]
Manik-Jain: https://github.com/eea-oasis/baseline/issues/593

Manik-Jain:
> Some general comments:
> 
> 1. Please follow the branch naming standard of the Baseline repository.
> 2. Please provide more meaningful commit messages. General agreement is to do "This commit does..." approach where your commit message just continues the sentence. Also, commits should be much smaller.
> 3. Please follow the process we agreed on in the beginning of the project for defining the implementation direction.
>    ```
>        Create a doc with proposal -> Iterate with the team -> Write down final agreement -> Create separate smaller issues for implementation -> Implement by doing a single PR per issue
>    ```
> 
> This is important for a multitude of reasons, most notably:
> 
> * We document our design decisions in a transparent and consistent way
> * We do most of the design during the initial iterations by the team, avoiding costly rewrites and discussions once PR is created
> * We produce smaller PRs, easier to review thus with higher quality
> 
> 4. Please follow the agreed architecture of the project - I believe there is no need to explain why.
> 
> As we are limited with resources and are trying to fit a lot of work til the end of the year, following the process is essential for keeping things on time and producing high quality output.

1. Applies to others also. 
2. Can you quote the wrong commit message in this PR? 
3. **Document, iterations, agreements**: https://docs.google.com/document/d/1A4wr7Vqt9mUoDezKI5njk6eKRixGP1nUgxUe_E7pay0/edit

**Issues** : https://github.com/eea-oasis/baseline/issues/590
https://github.com/eea-oasis/baseline/issues/589
https://github.com/eea-oasis/baseline/issues/588

Since most of them need to be tested together, it was not pragmatic to segregate and test

4. I agree CQRS was missed as this whole PR was implemented in short span of time

I think we can discuss more courteously to collaborate more to SRI.

examples/bri-3/src/bri/auth/api/auth.controller.ts
Manik-Jain: latest commit has the nonce storage changes

examples/bri-3/src/bri/auth/strategy/jwt.strategy.ts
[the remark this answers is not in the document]
Therecanbeonlyone1969: @ognjenkurtic with all due respect, it is not, because it must be resolvable by **anyone**, and even **outside** of the BPI context since e.g. messaging between participants will be outside of the BPI context but using BpiSubject identifiers. See https://github.com/eea-oasis/baseline-standard/blob/main/core/baseline-core-v1.0-psd01.md#r41 and https://github.com/eea-oasis/baseline-standard/blob/main/core/baseline-core-v1.0-psd01.md#r42 ... we cannot assume that the BPI will be directly available for resolution of identifiers in the context of inter-party messaging.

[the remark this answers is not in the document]
Therecanbeonlyone1969: @ognjenkurtic thank you for the clarification. your comment makes sense now.

Therecanbeonlyone1969:
@Manik-Jain @ognjenkurtic @skosito @Kasshern @Ybittan what are we doing with this now?

ognjenkurtic:
@Manik-Jain Would you like someone of us to pick up and finish this one?

ognjenkurtic:
> @Manik-Jain @ognjenkurtic @skosito @Kasshern @Ybittan what are we doing with this now?

Either @Manik-Jain or someone from the group will finish this one.

examples/bri-3/front-end/README.md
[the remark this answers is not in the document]
skosito: let's just keep create-react-app, it is simple boilerplate anyways, the only thing is it should've just been separate PR so we could merge it

i understand simplicity argument, but create-react-app is simple boilerplate already, and super extensible for future... if we decided already to have UI to use for demo, we will probably have the need for couple of components down the line, let's at least make it look normal with some material UI libraries

examples/bri-3/front-end/src/login/Login.js
```diff
@@ -0,0 +1,36 @@
+import axios from 'axios';
+import { SRI_BACKEND } from '../constants';
+
+const handleLogin = async () => {
+  if (window.ethereum) {
+    window.ethereum.enable();
```
Kasshern, line 6: I got this warning in the browser console : MetaMask: 'ethereum.enable()' is deprecated and may be removed in the future. Please use the 'eth_requestAccounts' RPC method instead.
For more information, see: https://eips.ethereum.org/EIPS/eip-1102

examples/bri-3/front-end/src/login/Login.js
```diff
@@ -0,0 +1,36 @@
...
+    const accounts = await window.ethereum.send('eth_requestAccounts');
```
Kasshern, line 7: Also got this warning in the browser console: MetaMask: 'ethereum.send(...)' is deprecated and may be removed in the future. Please use 'ethereum.sendAsync(...)' or 'ethereum.request(...)' instead.
For more information, see: https://eips.ethereum.org/EIPS/eip-1193

Kasshern (changes requested):
Left some comments mainly on the frontend setup seeing as some large conversations already took place or are taking place on the nestjs side. Some cleanup of react images floating around could be done but I wont die on that hill.

skosito (dismissed):
all fixes mentioned in comments are done in this or follow up PRs, approving this one

Kasshern (dismissed):
All concerns addressed in Clean-up PRs

skosito:
@ognjenkurtic can you remove request for change since it is blocking the merge, if there are no more comments from your side?

skosito (approved):

ognjenkurtic (approved):

PR #606, opened 2022-12-12
Sri did auth

skosito:
# Description

Added create and verify jwt using did-jwt and resolver, check out this 2:
https://www.npmjs.com/package/did-jwt
https://www.npmjs.com/package/ethr-did-resolver

Ended up not using veramo since it seems that integrating their whole framework and concepts into nestjs architecture is overhead. Checking their source code related to this it seems it is a wrapper around these couple libraries anyways. I think keeping it simple to start until we need some plugin architecture like veramo should be a way to go first, but no strong opinion.

Did auth (includes nodejs express repo that uses above packages):
[https://rsksmart.github.io/rif-identity-docs/ssi/specs/did-auth.html](https://www.google.com/url?q=https://rsksmart.github.io/rif-identity-docs/ssi/specs/did-auth.html&sa=D&source=docs&ust=1670811168930873&usg=AOvVaw3fASjzIwlFCVfX3s_mfjal)

In our case challenge is generated uuid, which gets signed by metamask. Then signature and ethereum address are passed to auth endpoint and jwt is created using did-jwt. I am just using `did:ethr:0x5` here which is resolved on goerli (0x5 is goerli network).
Also note some env vars to add locally for this to work, check env.sample.

If this PR makes sense I will do additional cleanup and we can merge 3 auth PRs we have already, this one is branched out from one of them.

## Related Issue

Check base PR.

## Motivation and Context

Check base PR.

## How Has This Been Tested

Use frontend to get jwt token in browser console, then copy and paste it to postman collection access_token variable.

## Types of changes

## Checklist


examples/bri-3/src/bri/auth/capabilities/login/login.command.handler.ts
```diff
@@ -24,10 +24,18 @@ export class LoginCommandHandler implements ICommandHandler<LoginCommand> {
       throw new Error(errorMessage.USER_NOT_AUTHORIZED);
     }
 
-    const jwtPayload = {
-      username: bpiSubject.name,
-      sub: bpiSubject.id,
+    const serviceUrl = process.env.SERVICE_URL;
+    const subjectDid = `did:ethr:0x5:${publicKey}`;
```
skosito, line 28: not the best code, just for illustration - should we send did or public key in login request? both should work i think

examples/bri-3/src/bri/auth/guards/didJwt.guard.ts
```diff
@@ -0,0 +1,73 @@
+import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
+import { Reflector } from '@nestjs/core';
+import { IS_PUBLIC_ENDPOINT_KEY } from 'src/bri/decorators/public-endpoint';
+import { Resolver } from 'did-resolver';
+import { getResolver } from 'ethr-did-resolver';
+import { verifyJWT } from 'did-jwt';
+
+@Injectable()
+export class DidJwtAuthGuard implements CanActivate {
```
skosito, line 9: custom guard not extending passport guards

examples/bri-3/src/bri/auth/capabilities/login/login.command.handler.ts
Therecanbeonlyone1969: did does not change ever, pub keys do. So let's use the did since that should be the userId in the bpiSubjectAccount

Therecanbeonlyone1969 (approved):
LGTM

examples/bri-3/src/bri/auth/agent/auth.agent.ts
```diff
@@ -29,4 +30,17 @@ export class AuthAgent {
     const access_token = this.jwtService.sign(payload);
     return { access_token };
   }
+
+  async generateDidJwt(payload: any) {
+    const serviceDid = process.env.GOERLI_SERVICE_DID;
+    const privateKey = process.env.GEORLI_SERVICE_SIGNER_PRIVATE_KEY;
+    const serviceSigner = ES256KSigner(hexToBytes(privateKey));
+    const access_token = await createJWT(
```
Manik-Jain, line 38: would be good to handle the cases if env variables are either empty or undefined

examples/bri-3/src/bri/auth/guards/didJwt.guard.ts
```diff
@@ -0,0 +1,73 @@
...
+      console.log('jwt verification error ', e);
+      return false;
+    }
+  }
+
+  private async getDidResolver() {
+    const rpcUrl = process.env.GOERLI_RPC_URL;
+
+    const providerConfig = {
+      networks: [{ name: '0x5', rpcUrl }],
```
Manik-Jain, line 36: would be better to move to a constant in order to save efforts if this changes anytime down the line

Manik-Jain:
there are some console logs in this change., do we really need them?

examples/bri-3/src/bri/auth/agent/auth.agent.ts
skosito: we can, but since they pretty much must be defined, and cant be defaulted (for example private key can not have fallback value) all methods using them will throw exception and methods calling this will catch
is there some way you suggest we handle this?

Manik-Jain: currently we are not assigning the field types to string, and hence TS doesn't complain.
But, if you do something like `const serviceDid : string = process.env.GOERLI_SERVICE_DID;` TS may complain that the value should be a string.

And yes, you have rightly mentioned that any functions using the undefined values will throw exception, that's why we should handle them all.

skosito: let's discuss error handling in separate issue, current i see bunch of agents can throw exceptions but i dont see how we are catching them in command handlers
cc: @ognjenkurtic i forgot but i think we discussed this

examples/bri-3/src/bri/auth/capabilities/login/login.command.handler.ts
skosito: @Therecanbeonlyone1969 could you please clarify this? currently bpiSubject contains publicKey and we are fetching bpiSubject here by public key - not sure i follow, should did replace publicKey in bpiSubject or something else? also there is no userId in bpiSubjectAccount, there are db foreign key relations with bpiSubject (ownerBpiSubjectId and creatorBpiSubjectId)

if we need some db changes too let's please do it in separate issue and merge current auth work so we dont have to maintain 3 PRs, we can build up on top of this

Kasshern (approved):

Issue #604, opened 2022-12-01
[SRI] Workgroup Setup Functionality


Kasshern:
## Overview

Create SRI Workgroup Setup Functionality within the BRI-3 project utilizing the Workgroup Setup Proposal document as a guide

### Reference

- [Workgroup Setup Proposal ](https://docs.google.com/document/d/1pV9yOIRKfj9uBobKh1s2KZI44gfmWOEDr1_wTqtxPZk/edit?usp=sharing)

### Assumptions

- Authentication, Authorization, Workgroup Invitation, and any other code necessary to fully complete this functionality, but is not included within the scope of this issue will be commented as TODO

### Acceptance

- Alter CreateWorkgroup method in accordance with available auth
- Implement creator as administrator and participant 
- Implement default starting privacy and security policies
- Assumes initial state between participants
- Implement Unit tests assessing Workgroup Component above implementations

## Tasks

- [x] Alter CreateWorkgroup method to default add creator as admin and participant
- [x] Alter default sec and priv policies in accordance with new sec and priv policies
- [x] Implement Unit tests assessing Workgroup Setup functionality

PR #614, opened 2023-01-13
Feature/604 bri3 workgroup setup

Kasshern:
# Description

Alters workgroup creation endpoint to create a workgroup that by default adds the creator as the admin and participant through public key extracted from JSON web token. Alters related tests, DTOs, etc. to reflect new method of workgroup setup. 

## Related Issue
[Workgroup Setup Functionality](https://github.com/eea-oasis/baseline/issues/604)

## Motivation and Context

Authenticated parties should be able to setup a workgroup and have themselves added to it immediately without sending extra information beyond their access token. 

## How Has This Been Tested

Project test suites  run with 100% pass

## Types of changes

## Checklist

### DIFF
--- a/examples/bri-3/src/bri/decorators/public-key.ts
+++ b/examples/bri-3/src/bri/decorators/public-key.ts
@@ -1,0 +1,12 @@
+import { createParamDecorator, ExecutionContext } from '@nestjs/common';
+import { decodeJWT } from 'did-jwt';
+
+export const PublicKey = createParamDecorator(
+  (data: unknown, ctx: ExecutionContext) => {
+    const request = ctx.switchToHttp().getRequest();
+    const bearerToken = request.headers['authorization'];
+    const token = bearerToken.split(' ')[1];
+    const publicKey = decodeJWT(token).payload.sub.split(':')[3];
+    return publicKey;
+  },
+);
--- a/examples/bri-3/src/bri/decorators/public-key.ts
+++ b/examples/bri-3/src/bri/decorators/public-key.ts
@@ -1,10 +1,18 @@
-import { createParamDecorator, ExecutionContext } from '@nestjs/common';
+import {
+  createParamDecorator,
+  ExecutionContext,
+  NotFoundException,
+} from '@nestjs/common';
 import { decodeJWT } from 'did-jwt';
+import { TOKEN_NOT_FOUND } from '../shared/constants';
 
 export const PublicKey = createParamDecorator(
   (data: unknown, ctx: ExecutionContext) => {
     const request = ctx.switchToHttp().getRequest();
     const bearerToken = request.headers['authorization'];
+    if (!bearerToken) {
+      throw new NotFoundException(TOKEN_NOT_FOUND);
+    }
     const token = bearerToken.split(' ')[1];
     const publicKey = decodeJWT(token).payload.sub.split(':')[3];
     return publicKey;
--- a/examples/bri-3/src/bri/decorators/public-key.ts
+++ b/examples/bri-3/src/bri/decorators/public-key.ts
@@ -6,6 +6,7 @@ import {
 import { decodeJWT } from 'did-jwt';
 import { TOKEN_NOT_FOUND } from '../shared/constants';
 
+//TODO implement command using request-context - retrieve bpiSubject by publicKey
 export const PublicKey = createParamDecorator(
   (data: unknown, ctx: ExecutionContext) => {
     const request = ctx.switchToHttp().getRequest();
--- a/examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
+++ b/examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
@@ -51,6 +51,20 @@ export class BpiSubjectAgent {
     return bpiSubjectToUpdate;
   }
 
+  public async fetchUpdateCandidatesAndThrowIfNoneExist(
+    ids: string[],
+  ): Promise<BpiSubject[]> {
+    const bpiSubjects = await this.storageAgent.getBpiSubjectsById(ids);
+
+    bpiSubjects.forEach((bp) => {
+      if (!ids.includes(bp.id)) {
+        throw new NotFoundException(NOT_FOUND_ERR_MESSAGE);
+      }
+    });
+
+    return bpiSubjects;
+  }
+
   public updateBpiSubject(
     bpiSubjectToUpdate: BpiSubject,
     name: string,
--- a/examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
+++ b/examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
@@ -51,16 +51,14 @@ export class BpiSubjectAgent {
     return bpiSubjectToUpdate;
   }
 
-  public async fetchUpdateCandidatesAndThrowIfNoneExist(
+  public async fetchUpdateCandidatesAndThrowIfValidationFails(
     ids: string[],
   ): Promise<BpiSubject[]> {
     const bpiSubjects = await this.storageAgent.getBpiSubjectsById(ids);
 
-    bpiSubjects.forEach((bp) => {
-      if (!ids.includes(bp.id)) {
-        throw new NotFoundException(NOT_FOUND_ERR_MESSAGE);
-      }
-    });
+    if (ids.length != bpiSubjects.length) {
+      throw new NotFoundException(NOT_FOUND_ERR_MESSAGE);
+    }
 
     return bpiSubjects;
   }
--- a/examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
+++ b/examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
@@ -74,4 +74,18 @@ export class BpiSubjectAgent {
 
     return bpiSubjectToDelete;
   }
+
+  public async fetchBpiSubjectsByIdAndThrowIfNoneExist(
+    bpiSubjectIds: string[],
+  ): Promise<BpiSubject[]> {
+    const bpiSubjects = await this.storageAgent.getBpiSubjectsById(
+      bpiSubjectIds,
+    );
+
+    if (bpiSubjectIds.length !== bpiSubjects.length) {
+      throw new NotFoundException(NOT_FOUND_ERR_MESSAGE);
+    }
+
+    return bpiSubjects;
+  }
 }
--- a/examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
+++ b/examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
@@ -6,7 +6,6 @@ import {
 import { v4 } from 'uuid';
 import { BpiSubject } from '../models/bpiSubject';
 import { BpiSubjectType } from '../models/bpiSubjectType.enum';
-
 import {
   NAME_EMPTY_ERR_MESSAGE,
   NOT_FOUND_ERR_MESSAGE,
@@ -73,19 +72,5 @@ export class BpiSubjectAgent {
     }
 
     return bpiSubjectToDelete;
-  }
-
-  public async fetchBpiSubjectsByIdAndThrowIfNoneExist(
-    bpiSubjectIds: string[],
-  ): Promise<BpiSubject[]> {
-    const bpiSubjects = await this.storageAgent.getBpiSubjectsById(
-      bpiSubjectIds,
-    );
-
-    if (bpiSubjectIds.length !== bpiSubjects.length) {
-      throw new NotFoundException(NOT_FOUND_ERR_MESSAGE);
-    }
-
-    return bpiSubjects;
   }
 }
--- a/examples/bri-3/src/bri/workgroup/workgroups/agents/workgroups.agent.ts
+++ b/examples/bri-3/src/bri/workgroup/workgroups/agents/workgroups.agent.ts
@@ -5,53 +5,13 @@ import { Workstep } from '../../worksteps/models/workstep';
 
 import { uuid } from 'uuidv4';
 import { Workgroup } from '../models/workgroup';
-import { BpiSubjectStorageAgent } from '../../../identity/bpiSubjects/agents/bpiSubjectsStorage.agent';
 import { WorkgroupStorageAgent } from './workgroupStorage.agent';
 import { WORKGROUP_NOT_FOUND_ERR_MESSAGE } from '../api/err.messages';
-import { NOT_FOUND_ERR_MESSAGE as BPISUBJECT_NOT_FOUND_ERR_MESSAGE } from '../../../identity/bpiSubjects/api/err.messages';
 
 // Agent methods have extremely declarative names and perform a single task
 @Injectable()
 export class WorkgroupAgent {
-  constructor(
-    private workgroupStorageAgent: WorkgroupStorageAgent,
-    private bpiSubjectStorageAgent: BpiSubjectStorageAgent,
-  ) {}
-
-  public async fetchBpiSubjectsByIdAndThrowIfNoneExist(
-    bpiSubjectIds: string[],
-  ): Promise<BpiSubject[]> {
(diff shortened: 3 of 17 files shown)


examples/bri-3/src/bri/workgroup/workgroups/api/workgroups.controller.spec.ts
```diff
@@ -23,16 +23,16 @@ import { classes } from '@automapper/classes';
 
 describe('WorkgroupsController', () => {
   let workgroupController: WorkgroupController;
-
   let mockBpiSubjectStorageAgent: MockBpiSubjectStorageAgent;
   let mockWorkgroupStorageAgent: MockWorkgroupStorageAgent;
 
+  const tokenHeader =
```
Therecanbeonlyone1969, line 29: @Kasshern presume the hard coded values are only there for testing and until the DB runs. Two things:

- please, either mark as a ToDo wherever there is a hard-coded value to be moved into either a separate test setup or a configuration file
- or move the hardcoded values into a configuration file that is loaded into the module

examples/bri-3/src/bri/workgroup/workgroups/api/workgroups.controller.spec.ts
```diff
@@ -43,17 +43,16 @@ describe('WorkgroupsController', () => {
       'name',
       'desc',
       BpiSubjectType.External,
-      'publicKey',
+      '0x9a36744bCD7fb5b41550e4719FFeA8B1B8c6A7bE',
```
Therecanbeonlyone1969, line 46: see above

Therecanbeonlyone1969 (approved):
See my two comments. good to merge once addressed @Kasshern

skosito (commented):
looks good, left couple of comments and suggestions

examples/bri-3/src/bri/workgroup/workgroups/api/workgroups.controller.spec.ts
Kasshern: Although this file has been altered these hardcoded values are used to instantiate entities within mock storage. Each controller spec tests uses hardcoded entities for mocks of storage.

examples/bri-3/src/bri/workgroup/workgroups/api/workgroups.controller.spec.ts
Kasshern: Although this file has been altered these hardcoded values are used to instantiate entities within mock storage. Each controller spec tests uses hardcoded entities for mocks of storage.

examples/bri-3/src/bri/workgroup/workgroups/agents/workgroups.agent.ts
```diff
@@ -18,44 +18,39 @@ export class WorkgroupAgent {
     private bpiSubjectStorageAgent: BpiSubjectStorageAgent,
   ) {}
 
-  public async fetchWorkgroupAdministratorsAndThrowIfNoneExist(
-    administratorIds: string[],
+  public async fetchBpiSubjectsByIdAndThrowIfNoneExist(
```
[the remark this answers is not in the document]
Kasshern: moved this specific fetch to bpisubjectsagent.

examples/bri-3/src/bri/workgroup/workgroups/agents/workgroups.agent.ts
```diff
@@ -18,44 +18,39 @@ export class WorkgroupAgent {
...
     return bpiSubjects;
   }
 
-  public async fetchWorkgroupParticipantsAndThrowIfNoneExist(
-    participantIds: string[],
-  ): Promise<BpiSubject[]> {
-    const bpiSubjects = await this.bpiSubjectStorageAgent.getBpiSubjectsById(
-      participantIds,
-    );
+  public async fetchBpiSubjectByPublicKeyAndThrowIfNoneExist(
```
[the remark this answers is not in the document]
Kasshern: replaced this with bpiSubjectStorage call

examples/bri-3/src/bri/workgroup/workgroups/api/workgroups.controller.ts
```diff
@@ -28,14 +29,14 @@ export class WorkgroupController {
   @Post()
   async createWorkgroup(
     @Body() requestDto: CreateWorkgroupDto,
+    @Headers('authorization') accessToken: string, //TODO replace with JWT type from its module
   ): Promise<string> {
     return await this.commandBus.execute(
       new CreateWorkgroupCommand(
+        accessToken,
```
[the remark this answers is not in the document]
Kasshern: I created a custom decorator in the shared folder containing the public endpoint decorator that takes in the jwt from the header, decodes it and passes back the pubkey.

examples/bri-3/src/bri/workgroup/workgroups/api/workgroups.controller.ts
```diff
@@ -28,14 +29,14 @@ export class WorkgroupController {
...
         requestDto.name,
-        requestDto.administratorIds,
-        requestDto.securityPolicy,
-        requestDto.privacyPolicy,
-        requestDto.participantIds,
+        requestDto.securityPolicy, //TODO will be made default once secPolicy Implemented
+        requestDto.privacyPolicy, //TODO will be made default once priPolicy Implemented
```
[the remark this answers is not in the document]
Kasshern: replaced with comments to existing issues. The default thing was me just making a note that upon workgroup creation each workgroup would by default have certain policies like it does for admin and participant

examples/bri-3/src/bri/decorators/public-key.ts
```diff
@@ -0,0 +1,12 @@
+import { createParamDecorator, ExecutionContext } from '@nestjs/common';
+import { decodeJWT } from 'did-jwt';
+
+export const PublicKey = createParamDecorator(
+  (data: unknown, ctx: ExecutionContext) => {
+    const request = ctx.switchToHttp().getRequest();
+    const bearerToken = request.headers['authorization'];
+    const token = bearerToken.split(' ')[1];
+    const publicKey = decodeJWT(token).payload.sub.split(':')[3];
```
Manik-Jain, line 9: can token be undefined in worst case?

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
```diff
@@ -74,4 +74,18 @@ export class BpiSubjectAgent {
...
   }
+
+  public async fetchBpiSubjectsByIdAndThrowIfNoneExist(
+    bpiSubjectIds: string[],
+  ): Promise<BpiSubject[]> {
+    const bpiSubjects = await this.storageAgent.getBpiSubjectsById(
+      bpiSubjectIds,
+    );
+
+    if (bpiSubjectIds.length !== bpiSubjects.length) {
```
Manik-Jain, line 85: not sure if we shall throw an exception, or log the ids which have data. 
@Therecanbeonlyone1969 what would you suggest?

Therecanbeonlyone1969 (approved):
LGTM ... hardcoded values should still be marked as toDo to move into config files

examples/bri-3/src/bri/decorators/public-key.ts
Kasshern: @Manik-Jain I don't see a way for the token to be undefined if this decorator is used with a guarded route, but will add a thrown error just in the decorator is used improperly.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
Kasshern: @Manik-Jain I will check out the behavior of the findMany used on the ORM side. Assuming that it will return a result as long as at least one record is found. If that is the case I think those found can be sent back to here in the code and I filter out the ones found and return them while logging the ones not found with an appropriate message. @skosito @ognjenkurtic @Ybittan what do you think?

Therecanbeonlyone1969:
> > LGTM ... hardcoded values should still be marked as toDo to move into config files
> 
> @Therecanbeonlyone1969 I can make an issue for adding TODOs for every hardcoded values within the tests that do mock storage setup, but that will essentially be every controller spec.

Please, do @Kasshern

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
Ybittan: On the assumption that it will return results if there is a record found, I would agree filtering them out to return here and logging ones not found makes sense to me as a solution to ensure the right data goes through, while still properly logging what is missing.

Manik-Jain: right

examples/bri-3/src/bri/workgroup/workgroups/api/workgroups.controller.spec.ts
```diff
@@ -43,17 +40,16 @@ describe('WorkgroupsController', () => {
       'name',
       'desc',
       BpiSubjectType.External,
-      'publicKey',
+      'pubkey',
```
Manik-Jain, line 43: sorry, any specific reasons to rename this?

Kasshern: no specific reason.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
Kasshern: Removed this functionality as its not being used anymore. will be added back in when invitation is done. @Manik-Jain @Ybittan

examples/bri-3/src/bri/decorators/public-key.ts
Kasshern: I ended up doing the check on bearer token in case someone places this decorator on a public path. that is the only way they will reach this point without an authorization header. If the bearer token exists (even in an incorrect form) split should still return something defined even if delimiter is not found. Incorrect forms will be caught on the next line within the decodeJWT when identified as incorrect format.

examples/bri-3/src/bri/workgroup/workgroups/api/workgroups.controller.ts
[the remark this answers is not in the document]
ognjenkurtic: Yes, passing access token to every command that needs to be aware of the currently logged in user is probably not the optimal option. We could use something like this https://www.npmjs.com/package/request-context and then have a wrapper around it so that we can fetch the bpi subject id wherever needed. Would you mind adding a TODO and opening an issue to discuss further?

examples/bri-3/src/bri/workgroup/workgroups/capabilities/createWorkgroup/createWorkgroupCommand.handler.ts
```diff
@@ -13,22 +14,18 @@ export class CreateWorkgroupCommandHandler
   ) {}
 
   async execute(command: CreateWorkgroupCommand) {
-    const workgroupAdministrators =
-      await this.agent.fetchWorkgroupAdministratorsAndThrowIfNoneExist(
-        command.administratorIds,
-      );
+    const token = command.accessToken.split(' ')[1];
+    const publicKey = decodeJWT(token).payload.sub.split(':')[3];
```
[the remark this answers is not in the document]
ognjenkurtic: Look at my comment above related to passing the token via command. The code here should look something like this by using request context lib:

```
const currentUserPk = this.authService.getCurrentUserPublicKey();
const workgroupCreator =
      await this.agent.fetchBpiSubjectByPublicKeyAndThrowIfNoneExist(currentUserPk);
```

or like this if we introduce BaseCommandHandler:

```
const workgroupCreator =
      await this.agent.fetchBpiSubjectByPublicKeyAndThrowIfNoneExist(CurrentUserPk);
```
Please open an issue and add a TODO

examples/bri-3/src/bri/workgroup/workgroups/capabilities/updateWorkgroup/updateWorkgroupCommand.handler.ts
```diff
@@ -9,17 +10,18 @@ export class UpdateWorkgroupCommandHandler
...
-    private storageAgent: WorkgroupStorageAgent,
+    private workgroupStorageAgent: WorkgroupStorageAgent,
+    private bpiSubjectStorageAgent: BpiSubjectStorageAgent,
   ) {}
 
   async execute(command: UpdateWorkgroupCommand) {
     const adminstratorsToUpdate =
-      await this.agent.fetchWorkgroupAdministratorsAndThrowIfNoneExist(
+      await this.bpiSubjectStorageAgent.getBpiSubjectsById(
         command.administratorIds,
```
ognjenkurtic, line 20: Since both adminstratorIds and participantIds are user inputs, i would do a check here to make sure they are all present before updating the workgroup and would otherwise throw. For example, If we swallow the wrong Id the user might be unaware that the new admin has not been added to the group which could lead to a problem if the decides to ditch the old admin identity for whatever reason.

examples/bri-3/src/bri/decorators/public-key.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  createParamDecorator,
+  ExecutionContext,
+  NotFoundException,
+} from '@nestjs/common';
+import { decodeJWT } from 'did-jwt';
+import { TOKEN_NOT_FOUND } from '../shared/constants';
+
+export const PublicKey = createParamDecorator(
```
ognjenkurtic, line 9: See my comments on accessing the context is a separate service exposing a simple method to fetch the logged in user information.

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
```diff
@@ -51,6 +51,20 @@ export class BpiSubjectAgent {
     return bpiSubjectToUpdate;
   }
 
+  public async fetchUpdateCandidatesAndThrowIfNoneExist(
+    ids: string[],
+  ): Promise<BpiSubject[]> {
+    const bpiSubjects = await this.storageAgent.getBpiSubjectsById(ids);
+
+    bpiSubjects.forEach((bp) => {
```
ognjenkurtic, line 59: isn't it more efficient to just compare the length of both collections?

ognjenkurtic (dismissed):

examples/bri-3/src/bri/identity/bpiSubjects/agents/bpiSubjects.agent.ts
Kasshern: Yeah that's how it is on my local. Will push when I am back at computer. @ognjenkurtic

Kasshern: Done

examples/bri-3/src/bri/workgroup/workgroups/capabilities/updateWorkgroup/updateWorkgroupCommand.handler.ts
Kasshern: Implemented as length check

examples/bri-3/src/bri/decorators/public-key.ts
Kasshern: TODO added

examples/bri-3/src/bri/workgroup/workgroups/capabilities/createWorkgroup/createWorkgroupCommand.handler.ts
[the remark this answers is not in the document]
Kasshern: TODO added

examples/bri-3/src/bri/workgroup/workgroups/api/workgroups.controller.spec.ts
Kasshern: issue created

examples/bri-3/src/bri/workgroup/workgroups/api/workgroups.controller.spec.ts
Kasshern: Issue created

Therecanbeonlyone1969 (approved):
LGTM once all remaining comments are addressed

ognjenkurtic (approved):

Issue #620, opened 2023-01-19
Bugfix/Remove Confusing language in BRI-3 Readme


fleischr:
## Expected Behavior
All open source Baseline Protocol implementations are by their nature - vendor agnostic. 
Docs should focus on describing the BRI, not drawing incorrect or unhelpful comparisons between BRIs.

## Actual Behavior
The BRI-3 readme specifies itself as the "first vendor agnostic" implementation of the standard. This may mislead some to think they cannot run BRI-1/PRVD or other prior BRIs in a permissionless, open source way.
As differing implementations of Baseline can and will occur, identifying one or the other as the "basis for the Baseline SDK" is actually ambiguous

## Steps to Reproduce the Problem

View readme for BRI-3

## Specifications

Remove incorrect/unhelpful language

fleischr:
Closing issue. No actions were taken.

PR #621, opened 2023-01-19
#620 - correct some language in BRI-3 readme

fleischr:
# Description

BRI-3 readme contains some confusing / misleading language. This PR removes it.

## Related Issue

See issue #620 

## Motivation and Context

Keep documentation focused and clear

## How Has This Been Tested

Doc change only

## Types of changes

## Checklist

- [ ] My code follows the code style of this project.

### DIFF
--- a/examples/bri-3/README.md
+++ b/examples/bri-3/README.md
@@ -13,11 +13,7 @@ This reference implementation is being built from scratch and will contain a lim
 
  * Entry point for developers to get to know the protocol, and start contributing in an iterative and atomic fashion
 
- * Basis for the Baseline SDK
-
  * Set an example of how to build an open-source project under the Baseline protocol in regard to transparency, community participation and collaboration
-
- * First vendor-agnostic implementation of the protocol.
 
 ## Prerequisites
 


Ybittan:
Given that BRI1 has provide technology deeply embedded into it (example: "This implementation of the Baseline Protocol leverages the Provide stack for security"), and even has its readme titled "shuttle-app", is it really fair to call it vendor-agnostic?

skosito (changes requested):
yes, this is first vendor agnostic implementation, not developed by any company or tied to any company, what is incorrect there?

from your issue:
```
This may mislead some to think they cannot run BRI-1/PRVD or other prior BRIs in a permissionless, open source way.
```
Never in readme was mentioned that this is the only permissionless, open source implementation. 

not sure about basis for baseline SDK part, and if it's planned to build upon it, we can maybe change to "might be used as basis for baseline sdk" because there is no plan yet for sdk, but can double check

can be discussed on a core dev call and with bri-3 team, until then not approving

fleischr:
> Given that BRI1 has provide technology deeply embedded into it (example: "This implementation of the Baseline Protocol leverages the Provide stack for security"), and even has its readme titled "shuttle-app", is it really fair to call it vendor-agnostic?

BRI-1 was made vendor agnostic at the time its repos were open sourced and included to the PRVD Oasis charter. All aspects essential to running the BRI-1 implementation can be achieved independent to Provide Technologies Inc. In many cases, migration of the repos to PRVD has in fact required the removal of references to the Provide commercial infrastructure. There was significant effort and expense taken on to achieve this that ought be better understood. This is why I examine the claim of BRI-3 being the _first_ vendor-agnostic implementation of the standard debatable and misleading.

Overall, the term "vendor agnostic"  isn't a helpful distinction to make between one BRI to another when they are all open sourced in a similar manner. Particularly if it implies other BRIs *aren't* open sourced when in fact they are. This is why I recommend we stop using the term "vendor agnostic"

FYI -While Shuttle is a major enabler to BRI-1/PRVD, its not required.

Ybittan:
> FYI -While Shuttle is a major enabler to BRI-1/PRVD, its not required.

To me personally, as long as there is reference at all to Provide Technologies Inc. (or any other vendor) in the code and in the readme, it is not vendor agnostic. This is a helpful distinction for a new user, as there are pros and cons to utilizing reference material built by a vendor with their stack in mind. It's not a bad thing, and to the right user, it's a good thing, but it is something that needs to be made clear. 

Things like "FYI -While Shuttle is a major enabler to BRI-1/PRVD, its not required." is an immediate indication to me that it is not vendor agnostic. Out of the box, there should be no reference whatsoever to any commercial entity(ies) for it to be vendor agnostic. Again, in my opinion.

fleischr:
@Ybittan @skosito Disappointed that we can't be better aligned on this, but understand your objections.  Would still recommend more neutral language to be used here. If you think of different verbiage you'd like to use, I'd be happy to check it out later.

I'll close this PR for now to prevent the thread from going too far OT.

Issue #632, opened 2023-02-06
[SRI] Find a way to mock Prisma in way that allows us to ditch writing mock storage agents in unit tests


ognjenkurtic:
## Overview

This way we can mock the DB so that everything else is tested and we avoid writing bolierplate code.

### Reference

- [Prisma Client mock Github issue](https://github.com/prisma/prisma1/issues/3898)

### Questions

n/a

### Assumptions

- Nobody wants to write mock storage agents

### Acceptance

- Mock of prisma client implemented with a nice way to provide seed data
- All mock storage agents ditched
- All unit tests still pass

## Tasks

- [ ] Implenent it

Ybittan:
https://www.prisma.io/docs/guides/testing/unit-testing

PR #664, opened 2023-04-01
Feature/632 mock prisma client, remove transaction storage agent mock

skosito:
# Description

Demonstrates one mock approach on transactions storage agent since that is only used in transaction controller tests it was the easiest one to try it out.
We can not mock prisma client or prisma service it seems, because our storage agents are extending prisma service, but not injecting it through dependency injection, so we can mock storage agents instead (i think this also reduces boilerplate).

Two alternative approaches I could think of if this is not ok:
- inject prisma service in agents and mock prisma service, that way we can test logic in storage agents too, but i dont think its needed
- introduce prisma module and replace whole module with a mock, i think that way storage agents will extend mocked service (not 100% sure)

If this is ok, i will replace the rest of storage agents in follow up PR using same approach.

## Related Issue

https://github.com/eea-oasis/baseline/issues/632

## Motivation and Context

Removes the need for mock storage agents.

## How Has This Been Tested

Running transaction controller tests.

## Types of changes

## Checklist

- [] I have added tests to cover my changes.


Manik-Jain (approved):

examples/bri-3/src/bri/transactions/api/transactions.controller.spec.ts
```diff
@@ -106,41 +112,43 @@ describe('TransactionController', () => {
...
-        workstepInstanceId: '24',
-        fromSubjectAccountId: fromBpiSubjectAccount.id,
-        toSubjectAccountId: toBpiSubjectAccount.id,
-        payload: 'payload1',
-        signature: 'signature',
-      } as CreateTransactionDto;
-
-      const newTransactionId = await controller.createTransaction(requestDto);
+      const transaction = new Transaction(
+        '123',
```
ognjenkurtic, line 116: would it make sense we use named params here just for sake of better readability?

examples/bri-3/src/bri/transactions/api/transactions.controller.spec.ts
```diff
@@ -106,41 +112,43 @@ describe('TransactionController', () => {
...
+        'payload1',
+        'signature',
+        TransactionStatus.Initialized,
+      );
+      transactionStorageAgentMock.getTransactionById.mockResolvedValueOnce(
+        transaction,
+      );
 
       // Act
       const createdTransaction = await controller.getTransactionById(
```
ognjenkurtic, line 131: createdTransaction => fetchedTransaction?

examples/bri-3/src/bri/transactions/api/transactions.controller.spec.ts
skosito: there is an issue to move hardcoded test values in separate place, lets maybe clean up existing mock storage agents with this work because all other places have these hardcoded values, and then clean up naming and stuff?

ognjenkurtic (approved):

Kasshern (approved):

Issue #660, opened 2023-03-16
[SRI] Remove createAnchorHash API endpoint


biscuitdey:
## Overview

Remove createAnchorHash API endpoint in the controller as this function will only be called internally.

## Tasks

- [ ] Remove Api in controller
- [ ] Update createAnchorHash command handler

PR #667, opened 2023-04-16
Remove createAnchorHash API Endpoint

sf316:
# Description

Removed createAnchorHash API Endpoint. This is the minimum change to remove API Endpoint while keeping controller, command and command handler unchanged as well as passing existing tests locally.

Haven't update command and its handler. In the current model, commands and their handlers are only used in the controller, and the test for anchorHash controller is the only place called on the command and its handler. So do you still want to keep the controller, command and command handler since they can still be used internally? Or how would you like to update them?

## Related Issue

#660

## How Has This Been Tested

Run the app and check there is no more route for createAnchorHash.

image


Manik-Jain (approved):

ognjenkurtic:
@sf316 I think you can safely remove both the command and the command handler. Anchor hash creation will always happen as part of executing other commands, so there is no need for a specific command that only initiates anchor hash creation.

sf316:
> @sf316 I think you can safely remove both the command and the command handler. Anchor hash creation will always happen as part of executing other commands, so there is no need for a specific command that only initiates anchor hash creation.

I removed controller, command, command handler, the test as well as all imports or definitions associated with them. Is this what you want?

examples/bri-3/src/bri/zeroKnowledgeProof/api/anchorHash.controller.ts
```diff
@@ -1,24 +1,12 @@
 import { Body, Controller, Post } from '@nestjs/common';
 import { CommandBus } from '@nestjs/cqrs';
-import { CreateAnchorHashCommand } from '../capabilities/createAnchorHash/createAnchorHash.command';
 import { VerifyAnchorHashCommand } from '../capabilities/verifyAnchorHash/verifyAnchorHash.command';
-import { CreateAnchorHashDto } from './dtos/request/createAnchorHash.dto';
```
skosito, line 5: this dto can be removed too?

examples/bri-3/src/bri/zeroKnowledgeProof/api/anchorHash.controller.ts
```diff
@@ -1,24 +1,12 @@
...
 import { VerifyAnchorHashDto } from './dtos/request/verifyAnchorHash.dto';
-import { AnchorHashDto } from './dtos/response/anchorHash.dto';
```
skosito, line 7: is this dto used at some other place?

skosito (commented):
just one more question about potentially removing couple related dtos

examples/bri-3/src/bri/zeroKnowledgeProof/api/anchorHash.controller.ts
sf316: AnchorHashDto is also used in [anchorHash.profile.ts](https://github.com/eea-oasis/baseline/blob/main/examples/bri-3/src/bri/zeroKnowledgeProof/anchorHash.profile.ts)

examples/bri-3/src/bri/zeroKnowledgeProof/api/anchorHash.controller.ts
sf316: CreateAnchorHashDto is only used in createAnchorHash.dto.ts and its test. I have removed these two files and committed the change to this PR.

examples/bri-3/src/bri/zeroKnowledgeProof/api/anchorHash.controller.ts
ognjenkurtic: I believe it can also be removed as we used it only when creating a new anchor hash

sf316: @ognjenkurtic Do you mean remove AnchorHashDto completely as well? Or just remove AnchorHashDto in this controller?

Also I'm wondering do you need update this [doc](https://docs.google.com/document/d/1mkO5oou0TWk6kVHbTB01dyj7wuNGTfR9w6PAxSgXdGI/edit#) to reflect removing createAnchorHash? Thanks!

ognjenkurtic: Yes, you can remove AnchorHashDto completely from the codebase.The document you referenced is now out of date and the relevant one is the [SRI use-case](https://docs.google.com/document/d/14W2Pr00K8SBPPO7-aQiDP73uMAEwd6rwK7Tvtvfxeng/edit?usp=sharing) doc. We ll be updating all other documentation as soon as the discussion in the SRI use-case doc is finalized.

ognjenkurtic (approved):

examples/bri-3/src/bri/zeroKnowledgeProof/api/anchorHash.controller.ts
sf316: Done. Please let me know if there is anything else I can help with.

Kasshern (approved):

PR #678, opened 2023-05-08
Sparse Merkle Tree poc

ognjenkurtic:
# Description

Introduces a zk-kit sparse merkle tree package with a single test and a JSON representation of a tree from a merkelized invoices.

## Related Issue

n/a

## Motivation and Context

Opened in order for us to be able to discuss the SMT library and invoice merkelization. Does not have to be merged.

## How Has This Been Tested

n/a

## Types of changes

n/a

## Checklist

n/a


examples/bri-3/src/bri/storage/sparseMerkleTreePOC/smt.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    tree.add('4', '323032332d30362d3135'); // IssueDate:2023-06-15
+
+    tree.add('5', '323032332d30372d3135'); // DueDate:2023-07-15
+
+    tree.add('6', '4e4557'); // Status:NEW
+
+    // Invoice items are ignored for now
+
+    const anyTree = tree as any;
+    console.log(JSON.stringify(Array.from(anyTree.nodes)));
```
Manik-Jain, line 27: shouldn't that be an expect?

Issue #673, opened 2023-04-26
[SRI] - Extend the ProcessInboundMessageCommandHandler to allow for creation of transactions from BpiMessages


ognjenkurtic:
## Overview

We want to allow the user of the BPI to trigger a BpiTransaction by initiating a BpiMessage (i.e. via NATs). In order to do that, we need to extend the ProcessInboundMessageCommandHandler to trigger a tranasction creation in case BpiMessageType.Transaction.

### Reference

- [SRI use-case doc discusion](https://docs.google.com/document/d/14W2Pr00K8SBPPO7-aQiDP73uMAEwd6rwK7Tvtvfxeng/edit?disco=AAAApWHJw_Q)

### Questions

n/a

### Assumptions

if possible, we should just issue a CreateTransaction command if basic validation of the message succeeds. Transaction execution comes later down the line.

### Acceptance

- Extended the ProcessInboundMessageCommandHandler to check message type
- Triggered a CreateTransactionCommand if validation ok
- Written a test to verify behavior
- 
## Tasks

- [x] Implement it
- [x] Write tests

Issue #669, opened 2023-04-21
[SRI] Extend the BpiMessage model to allow for BpiTransaction triggering


ognjenkurtic:
## Overview

We want to allow the user of the BPI to trigger a BpiTransaction by initiating a BpiMessage (i.e. via NATs). In order to do that, we need to extend the BpiMessage model with relevant fields. Scope of this one is only to extend the BpiMessage (and maybe BpiTransaction, see bellow). We ll have a separate issue that deals with processing BpiMessages of this type.

### Reference

- [SRI use-case doc discusion 1](https://docs.google.com/document/d/14W2Pr00K8SBPPO7-aQiDP73uMAEwd6rwK7Tvtvfxeng/edit?disco=AAAApWHJw_Q)
- [SRI use-case doc discussion 2] (https://docs.google.com/document/d/14W2Pr00K8SBPPO7-aQiDP73uMAEwd6rwK7Tvtvfxeng/edit?disco=AAAAurnmVLw)

### Questions

- Are we all good with introducing a new BpiMessage type Transaction?
- BpiTransaction currently does not reference any specific Workgroup? Do we need to change that?

### Acceptance

- Extended the BpiMessageType enum with Transaction
- Extended the BpiMessage model with WokrgroupId, WorkflowId and WorkstepId to allow for specific workstep triggering when type Transaction
- Extend the BpiMessage model with BpiSubjectAccountId provided by the initiator when type transaction

## Tasks

- [x] Introduce BpiMessageType Transaction
- [ ] Extend the BpiMessage model as described

PR #684, opened 2023-05-15
Feature/673 extend processpnboundmessagecommandhandler to allow for transaction creation

ognjenkurtic:
# Description

Introduces BpiMessage Transaction type and additional infrastructure that enables triggering of a command to create a BpiTransaction from a NATS incoming message which is properly formatted. Contains a number of TODOs that will be tackled as part of #669 

## Related Issue

#673
#669 

## Motivation and Context

Standard defines that a BpiTransaction can be triggered through a BpiMessage send via a messaging system utilized by the BPI.

## How Has This Been Tested

Wrote new unit test and ran unit and e2e tests

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/examples/bri-3/src/bri/communication/agents/messaging.agent.spec.ts
+++ b/examples/bri-3/src/bri/communication/agents/messaging.agent.spec.ts
@@ -1,3 +1,4 @@
+import { BpiMessageType } from '../models/bpiMessageType.enum';
 import { MessagingAgent } from './messaging.agent';
 
 let messagingAgent: MessagingAgent;
@@ -96,7 +97,7 @@ describe('Messaging Agent', () => {
   it('Should return error when validating correct JSON raw message with message type not being known', () => {
     // Arrange
     const rawMessage =
-      '{ "id": "0a3dd67c-c031-4b50-95df-0bc5fc1c78b5", "fromBpiSubjectId": "71302cec-0a38-469a-a4e5-f58bdfc4ab32", "toBpiSubjectId": "76cdd901-d87d-4c87-b572-155afe45c128", "content": { "testProp":"testValue" }, "signature": "xyz", "type": 2}';
+      '{ "id": "0a3dd67c-c031-4b50-95df-0bc5fc1c78b5", "fromBpiSubjectId": "71302cec-0a38-469a-a4e5-f58bdfc4ab32", "toBpiSubjectId": "76cdd901-d87d-4c87-b572-155afe45c128", "content": { "testProp":"testValue" }, "signature": "xyz", "type": 999}';
 
     // Act
     const [, validationErrors] =
@@ -104,10 +105,10 @@ describe('Messaging Agent', () => {
 
     // Assert
     expect(validationErrors.length).toBe(1);
-    expect(validationErrors[0]).toEqual('type: 2 is unknown');
+    expect(validationErrors[0]).toEqual('type: 999 is unknown');
   });
 
-  it('Should return no errors and create message dto when validating correct JSON raw message', () => {
+  it('Should return no errors and create message dto when validating correct JSON raw message of INFO type', () => {
     // Arrange
     const rawMessage =
       '{ "id": "0a3dd67c-c031-4b50-95df-0bc5fc1c78b5", "fromBpiSubjectId": "71302cec-0a38-469a-a4e5-f58bdfc4ab32", "toBpiSubjectId": "76cdd901-d87d-4c87-b572-155afe45c128", "content": { "testProp":"testValue" }, "signature": "xyz", "type": 0}';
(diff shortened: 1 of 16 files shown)


biscuitdey (commented):
createNewBpiMessage() is used in two cases: 1. creating a new message & publishing it; 2. in the processInboundMessage command handler. In both cases, the function takes `id` as a param.

Case 1: Where is the uuid for the message (`id`) generated?

Case 2: In the processInboundMessage command handler, we are assuming that the other participant has NOT created a new BpiMessage before publishing and sending us the message. But in the createBpiMessage, a message can only be published after creating a new BpiMessage. Wouldn't this cause the processInboundMessage to exit in bpiMessageAlreadyExists?

In the processInboundMessage, I think we should check that the message we have received corresponds to an existing BpiMessage, and we are only handling the tasks or receiving the information of a message already created by some other participant.

examples/bri-3/src/bri/transactions/agents/transactions.agent.ts
```diff
@@ -17,11 +17,17 @@ import { BpiSubjectAccount } from '../../identity/bpiSubjectAccounts/models/bpiS
 @Injectable()
 export class TransactionAgent {
   constructor(private storageAgent: TransactionStorageAgent) {}
+
   public throwIfCreateTransactionInputInvalid() {
     // TODO: This is a placeholder, we will add validation rules as we move forward with business logic implementation
     return true;
   }
 
+  public isCreateTransactionInputInvalid(): boolean {
```
Kasshern, line 26: We need to be able to throw if the inpur is invalid and determine if the input is invalid without throwing?

examples/bri-3/src/bri/communication/agents/messaging.agent.spec.ts
```diff
@@ -128,6 +129,30 @@ describe('Messaging Agent', () => {
...
     expect(resultDto.content).toEqual({ testProp: 'testValue' });
     expect(resultDto.signature).toEqual('xyz');
-    expect(resultDto.type).toEqual(0);
+    expect(resultDto.type).toEqual(BpiMessageType.Info);
+  });
+
+  it('Should return no errors and create message dto when validating correct JSON raw message of TRANSACTION type', () => {
+    // Arrange
+    const rawMessage =
+      '{ "id": "0a3dd67c-c031-4b50-95df-0bc5fc1c78b5", "fromBpiSubjectId": "71302cec-0a38-469a-a4e5-f58bdfc4ab32", "toBpiSubjectId": "76cdd901-d87d-4c87-b572-155afe45c128", "content": { "testProp":"testValue" }, "signature": "xyz", "type": 1}';
```
Kasshern, line 138: Perhaps the enum type should be given string designations

Kasshern (dismissed):
Minor comments - Looks good overall

examples/bri-3/src/bri/communication/agents/messaging.agent.ts
```diff
@@ -72,12 +89,28 @@ export class MessagingAgent implements OnApplicationBootstrap {
...
       errors.push(`${rawMessage} is not valid JSON. Error: ${e}`);
       return [newBpiMessageCandidate, errors];
     }
 
+    if (
+      !newBpiMessageCandidate.isInfoMessage() &&
+      !newBpiMessageCandidate.isTransactionMessage()
+    ) {
+      errors.push(`type: ${newBpiMessageCandidate.type} is unknown`);
+      return [null, errors];
```
skosito, line 111: is there a reason to not just throw err or antything else, this seems a bit odd for ts

examples/bri-3/src/bri/communication/agents/messaging.agent.ts
```diff
@@ -53,16 +53,33 @@ export class MessagingAgent implements OnApplicationBootstrap {
...
-      new ProcessInboundBpiMessageCommand(
-        newBpiMessageCandidate.id,
-        newBpiMessageCandidate.fromBpiSubjectId,
-        newBpiMessageCandidate.toBpiSubjectId,
-        JSON.stringify(newBpiMessageCandidate.content),
-        newBpiMessageCandidate.signature,
-        newBpiMessageCandidate.type,
-      ),
-    );
+    if (newBpiMessageCandidate.isInfoMessage()) {
```
skosito, line 56: can this condition be checked in command handler since in both cases same command is called and that way we can isolate handling of different types on one place?

examples/bri-3/src/bri/identity/bpiSubjectAccounts/agents/bpiSubjectAccounts.agent.ts
```diff
@@ -17,20 +17,44 @@ export class BpiSubjectAccountAgent {
...
 
     return subjectAccounts;
   }
 
+  public async getBpiSubjectAccounts(
+    subjectAccountIds: string[],
+  ): Promise<BpiSubjectAccount[]> {
+    const subjectAccounts: BpiSubjectAccount[] = [];
+
+    for (const subjectAccountId of subjectAccountIds) {
```
skosito, line 44: nit but shouldn't this be one query instead of one query per id?

skosito (approved):
looks good, couple minor comments

ognjenkurtic:
> createNewBpiMessage() is used in two cases: 1. creating a new message & publishing it; 2. in the processInboundMessage command handler. In both cases, the function takes `id` as a param.
> 
> Case 1: Where is the uuid for the message (`id`) generated?
> 
> Case 2: In the processInboundMessage command handler, we are assuming that the other participant has NOT created a new BpiMessage before publishing and sending us the message. But in the createBpiMessage, a message can only be published after creating a new BpiMessage. Wouldn't this cause the processInboundMessage to exit in bpiMessageAlreadyExists?
> 
> In the processInboundMessage, I think we should check that the message we have received corresponds to an existing BpiMessage, and we are only handling the tasks or receiving the information of a message already created by some other participant.

Case 1:

Standard requirement is that the ID is produced by the sender. Sender generates one and passes it as part of the request DTO.
 [R112]
A BPI message MUST contain a unique message identifier that is generated by the sender.
This allows unique identification of the message through its lifecycle.

Case 2:

processInboundMessageCommandHandler deals with  the situation when the sender uses NATS to initiate message creation and publishing to the recipient

createBpiMessageCommandHandler deals with the situation when the senders does the same thing via the REST API. 

The logic in both is mostly the same - accept and parse input, create new message, publish message. Not sure what issue are you refering to. Let's walk through the code together on our next sync.

examples/bri-3/src/bri/transactions/agents/transactions.agent.ts
ognjenkurtic: Yes, we throw when we work with REST API and we do not throw when inside a background task (i.e. NATS message processing).

examples/bri-3/src/bri/communication/agents/messaging.agent.spec.ts
ognjenkurtic: You mean for better readability? We can open a separate issue for that but i would not tackle it as a priority. What do you think?

examples/bri-3/src/bri/communication/agents/messaging.agent.ts
ognjenkurtic: We want to collect all errors in one place and want to avoid throw as we are in a background task that should continue to execute

examples/bri-3/src/bri/communication/agents/messaging.agent.ts
ognjenkurtic: It is not the same command.

IsInfoMessage -> ProcessInboundBpiMessageCommand
isTransationMessage -> ProcessInboundBpiTransactionCommand

examples/bri-3/src/bri/identity/bpiSubjectAccounts/agents/bpiSubjectAccounts.agent.ts
ognjenkurtic: https://github.com/eea-oasis/baseline/issues/697

biscuitdey (approved):
Thanks for the explanation. LGTM.

examples/bri-3/src/bri/communication/agents/messaging.agent.spec.ts
Kasshern: not  priority

Kasshern (approved):

Issue #699, opened 2023-07-05
[SRI] Implement circuit from Workstep 1 from the SRI use-case docs


ognjenkurtic:
## Overview

Based on the description from the SRI use-case doc, we want to execute a circuit as part of workstep 1 execution that will verify that the rules related to invoice exchange have been followed. Scope of this story is to agree on the rules and implement that circuit.  

### Reference

- [SRI use-case doc](https://docs.google.com/document/d/14W2Pr00K8SBPPO7-aQiDP73uMAEwd6rwK7Tvtvfxeng/edit)

### Questions

- We need to agree on the exact rules and output of the circuit before this one is ready for implementation.

PR #706, opened 2023-07-06
Feature/600 sri zkp component implement zk circuit - Part 4

biscuitdey:
# Description

This PR adds snarkjs service and required tests.

## Related Issue

#699

## How Has This Been Tested

It has been tested using the accompanying unit tests.

## Types of changes

## Checklist


examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/snarkjs/snarkjs.service.ts
```diff
@@ -2,32 +2,71 @@ import { Injectable } from '@nestjs/common';
...
+      protocol: proof.protocol,
+      curve: proof.curve,
+    } as Proof;
+
+    return newProof;
   }
 
-  public async verifyProof(proof: Proof, witness: Witness): Promise<boolean> {
-    return true;
+  public async verifyProofUsingWitness(witness: Witness): Promise<boolean> {
```
Therecanbeonlyone1969, line 40: @biscuitdey you need to specify the protocol and curve for proof verification ... rn it is only groth16 on the default bn254 but that might change.

biscuitdey: The protocol and curve have been added as params inside the proof model

```
export class Proof {
  a: string[];
  b: string[][];
  c: string[];
  protocol?: string;
  curve?: string;
}
```

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/snarkjs/snarkjs.service.ts
```diff
@@ -2,32 +2,71 @@ import { Injectable } from '@nestjs/common';
...
-    return witness;
+  public witness: Witness;
+
+  public async createWitness(inputs: object): Promise<Witness> {
+    this.witness = new Witness();
+
+    const proof = await this.createProof(inputs);
+    this.witness.proof = proof;
+
+    const publicInputs = await this.getPublicInputs(inputs);
```
Therecanbeonlyone1969, line 19: @biscuitdey `getPublicInputs` runs `executeCircuit`, but why since `createProof` already executes the circuit. Should it not be simply:
```
const publicInputs = this.witness.proof.a
```
?

Therecanbeonlyone1969 (commented):
@biscuitdey See my questions/comments. Otherwise, it looks good.

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/snarkjs/snarkjs.service.ts
Therecanbeonlyone1969: @biscuitdey ah yes, we need to then check for different. We still need to check what the protocol and curve is and then choose the correct prover system and curve. So that needs to be added.

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/snarkjs/snarkjs.service.ts
biscuitdey: The snarkjs lib function for executing circuit returns the proof and public inputs together. I have added different functions to separate out these 2 values. Also, `this.witness.proof.a` is an element on the G1 elliptic curve. Public inputs values are different from this.

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/snarkjs/snarkjs.service.ts
biscuitdey: As we are using only groth16 for our simple use case, I think we can add the functionality for checking the different protocols and curves in a later PR.

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/snarkjs/snarkjs.service.ts
Therecanbeonlyone1969: yeah sorry a bit blind @biscuitdey 

can we not still do the following:
```
const { proof, publicInputs } = await this.executeCircuit(inputs);

const newProof = {
      a: proof.pi_a,
      b: proof.pi_b,
      c: proof.pi_c,
      protocol: proof.protocol,
      curve: proof.curve,
    };

this.witness.proof = newProof;
```

and remove `createProof` and `getPublicInputs` functions?

biscuitdey: sure I'll do that.

Therecanbeonlyone1969 (approved):
LGTM

Kasshern (approved):

Issue #668, opened 2023-04-21
[SRI] Add a reference to an BPI Account to a instance of a Workflow


ognjenkurtic:
## Overview

Based on the discussion in the SRI use-case doc, we agreed that the workgroup needs a way to store the state produced by a succesfull execution of a workstep to a Bpi Account. We agreed that the best way to achieve this is to add a reference to a BpiAccount to the instance of a Workflow in that Workgroup.

### Reference

- [SRI use-case doc discussion link](https://docs.google.com/document/d/14W2Pr00K8SBPPO7-aQiDP73uMAEwd6rwK7Tvtvfxeng/edit?disco=AAAApWHJw80)

### Questions

- Owners of the Bpi Account in question must be BpiSubjectAccounts from all BpiSubjects participating in the workgroup to which the workflow belongs?

### Assumptions

- For now, only the worksteps that are part of a workflow can change the state of its BpiAccount. This might change in the future.
- Once the workflow is created, no updates to its reference to the BpiAccount are allowed.

### Acceptance

- A reference to an existing BpiAccount is added to the workflow on workflow creation.
- All BpiSubjectAccount beloning to participants of the workgroup (BpiSubjects) are added as owners at the point of workflow creation
- 

## Tasks

- [ ] Implement descirbed behaviour
- [ ] Write unit tests to confirm it

PR #714, opened 2023-07-26
Feature/668 add reference to bpi account in a workflow

ognjenkurtic:
# Description

Introduces Bpi Account creation during creation of a Workflow, with a reference Workflow -> BpiAccount.

## Related Issue

#668 

## Motivation and Context

It allows for workflows to store state inside a BpiAccount which is co owned by all workgroup participants

## How Has This Been Tested

Partially by unit test. Full unit tests will come once we introduce https://www.prisma.io/docs/guides/testing/unit-testing

## Types of changes

## Checklist

### DIFF
--- a/examples/bri-3/src/bri/workgroup/workflows/models/workflow.ts
+++ b/examples/bri-3/src/bri/workgroup/workflows/models/workflow.ts
@@ -1,5 +1,6 @@
 import { AutoMap } from '@automapper/classes';
 import { BpiAccount } from '../../../identity/bpiAccounts/models/bpiAccount';
+import { Workgroup } from '../../workgroups/models/workgroup';
 import { Workstep } from '../../worksteps/models/workstep';
 
 export class Workflow {
@@ -14,6 +15,9 @@ export class Workflow {
 
   @AutoMap()
   workgroupId: string;
+
+  @AutoMap()
+  workgroup: Workgroup;
 
   @AutoMap()
   bpiAccountId: string;
@@ -32,6 +36,7 @@ export class Workflow {
     this.name = name;
     this.worksteps = worksteps;
     this.workgroupId = workgroupId;
+    this.bpiAccountId = bpiAccount.id;
     this.bpiAccount = bpiAccount;
   }
 
--- a/examples/bri-3/src/bri/workgroup/workflows/models/workflow.ts
+++ b/examples/bri-3/src/bri/workgroup/workflows/models/workflow.ts
@@ -15,6 +15,9 @@ export class Workflow {
   @AutoMap()
   workgroupId: string;
 
+  @AutoMap()
+  bpiAccountId: string;
+
   @AutoMap(() => BpiAccount)
   bpiAccount: BpiAccount;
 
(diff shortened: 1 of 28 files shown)


examples/bri-3/src/bri/workgroup/workflows/models/workflow.ts
```diff
@@ -14,16 +15,24 @@ export class Workflow {
...
+
+  @AutoMap(() => BpiAccount)
+  bpiAccount: BpiAccount;
+
   constructor(
     id: string,
     name: string,
     worksteps: Workstep[],
     workgroupId: string,
+    bpiAccount: BpiAccount,
```
Therecanbeonlyone1969, line 29: @ognjenkurtic should this not also contain the `bpiAccountId`?

examples/bri-3/src/bri/workgroup/workflows/models/workflow.ts
```diff
@@ -14,16 +15,24 @@ export class Workflow {
...
   ) {
     this.id = id;
     this.name = name;
     this.worksteps = worksteps;
     this.workgroupId = workgroupId;
+    this.bpiAccount = bpiAccount;
```
Therecanbeonlyone1969, line 35: same here

examples/bri-3/src/bri/workgroup/workgroups/agents/workgroupStorage.agent.ts
```diff
@@ -19,7 +19,7 @@ export class WorkgroupStorageAgent extends PrismaService {
         administrators: true,
         participants: true,
         workflows: {
-          include: { worksteps: true },
+          include: { worksteps: true, bpiAccount: true },
```
Therecanbeonlyone1969, line 22: `bpiAccountId`?

Therecanbeonlyone1969: slightly confused by the difference of bpiAccount which should be an object and bpiAccountId which should be a string

Kasshern (approved):

examples/bri-3/src/bri/workgroup/workflows/models/workflow.ts
ognjenkurtic: Good point. Usage of both bpiAccountId and bpiAccount (and this pattern is used across the codebase whenever there is a reference to an object) is not necessary. It is a convention we picked up from how prisma.schema objects are defined. I added a commit that updates bpiAccountId just in the ctor just to be consistent, will have to change this as part of a separate issue that i ll open now.

examples/bri-3/src/bri/workgroup/workgroups/agents/workgroupStorage.agent.ts
ognjenkurtic: Just a convention picked up from prisma.schema. I do not think we need both of these in the domain object (object reference should be enough) but i cant change that right now across the whole codebase. Let's just be consistent in how we do it and i ll open a separate issue to tackle it.

examples/bri-3/src/bri/workgroup/workflows/models/workflow.ts
ognjenkurtic: https://github.com/eea-oasis/baseline/issues/719

Therecanbeonlyone1969 (approved):
LGTM

Therecanbeonlyone1969 (approved):
LGTM

Issue #701, opened 2023-07-05
[SRI] Execute circuit as part of workstep 1 execution


ognjenkurtic:
## Overview

Based on the description from the SRI use-case doc, we need to execute a circuit as part of workstep 1 exection. Scope of this one is to prepare all inputs for the circuit execution and esecute it.

### Reference

- [SRI use-case doc](https://docs.google.com/document/d/14W2Pr00K8SBPPO7-aQiDP73uMAEwd6rwK7Tvtvfxeng/edit)

### Questions

- We need to agree on #698, #699, #700 before this one can be picked up

ognjenkurtic:
@biscuitdey is this something i could pick up?

biscuitdey:
@ognjenkurtic I am working on the PR to prepare the inputs for the circuit. You can start on executing the circuit inside workstep 1 using the `circuit.interface.ts` functions.

PR #720, opened 2023-08-07
Feature/701 execute circuit as part of workstep execution

ognjenkurtic:
# Description

This PR prepares everything for the execution of the circuit as part of transactionExecution in the transaction agent. It also introduces and documents a temporary convention by which circuits are connected to worksteps. It also introduces a transactionResult object which stores the results of the execution for future bpi account storage.

## Related Issue

#701 

## Motivation and Context

Sets up the stage for actually executing a circuit and storing the data.

## How Has This Been Tested

Manual test by running npm run test and npm run build. More extensive testing, both manual and automated, will follow.

## Types of changes

## Checklist


examples/bri-3/src/bri/transactions/agents/transactions.agent.ts
```diff
@@ -167,11 +174,74 @@ export class TransactionAgent {
...
+    );
+
+    return txResult;
+  }
+
+  // TODO: Only for the purposes of temporary convention
+  // to connect worksteps with circuits on the file system.
+  // Format is: <path_from_env>/<workstep_name_in_snake_case>_<predefined_suffix>.
+  // Will be ditchedcompletely as part of milestone 5.
+  private constructCircuitPathsFromWorkstepName(name: string): {
```
biscuitdey, line 202: These paths are autogenerated by snarkjs library when `npm run snarkjs:circuit {circuitName}` is executed. So the naming convention is determined by the snarkjs. We can pass `snakeCaseWorkstepName_circuit` as the `circuitName` when running `npm run snarkjs:circuit`, it will generate the files with the appropriate names.

examples/bri-3/.env.sample
```diff
@@ -16,5 +16,7 @@ BPI_NATS_SERVER_USER="bpi_operator"
 BPI_NATS_SERVER_PASS="liftboiliftboiliftboiliftboi1"
 BPI_ENCRYPTION_KEY_K_PARAM="yzkXp3vY_AZQ3YfLv9GMRTYkjUOpn9x18gPkoFvoUxQ"
 BPI_ENCRYPTION_KEY_KTY_PARAM="oct"
+SNARKJS_CIRCUIT_KEYS_PATH="zeroKnowledgeKeys/circuit/"
```
biscuitdey, line 19: I think we can remove these variables as the paths are autogenerated by the snarkjs library when the `npm run snarkjs:circuit {circuitName}` script is executed.

biscuitdey (changes requested):

examples/bri-3/.env.sample
ognjenkurtic: I was thinking in the direction that a workgroup admin would be required to provide a compiled circuit and all the necessary artifacts during the setup phase. They would generate those in their own environment and just place it in a predetermined folder where the BPI has access to it, and would have to follow a predefined naming convention.

This is why i think it makes sense to have the parh as part of configuration, because the BPI admin would be able to choose where to store the artifacts. If you agree, i would keep this config value with slight adjustments based on your inputs.

examples/bri-3/src/bri/transactions/agents/transactions.agent.ts
ognjenkurtic: I think this makes sense. We want all compiled cirucits and accompanying artifacts to follow a predefined convention - see comment below.

examples/bri-3/.env.sample
biscuitdey: Ok we can allow workgroup admins to provide compiled circuit and all the related artifacts. But for the demo, we would have to compile the circuits ourselves. For that purpose, we can use the code that I added.

biscuitdey (approved):
LGTM

Therecanbeonlyone1969 (approved):
LGTM

Issue #729, opened 2023-08-22
Extend E2E tests to cover milestone 4 use-case


ognjenkurtic:
## Overview

Extend the E2E test to cover the full flow from SRI use-case doc

### Reference

- [SRI use-case doc](https://docs.google.com/document/d/14W2Pr00K8SBPPO7-aQiDP73uMAEwd6rwK7Tvtvfxeng/edit)

### Acceptance

- Tests executes the full flow and verifies state storage via anchor  hash

## Tasks

- [ ] Implement the e2e test

Issue #702, opened 2023-07-05
[SRI] Update BpiAccount state after successful execution of the workstep


ognjenkurtic:
## Overview

Based on the description from the SRI use-case doc, we want to store outputs of the workstep execution in the shared BpiAccount state. Scope of this one is to agree on the final structure we want to use and then implement storage. 

### Reference

- [SRI use-case doc](https://docs.google.com/document/d/14W2Pr00K8SBPPO7-aQiDP73uMAEwd6rwK7Tvtvfxeng/edit)
- [State object Model proposal](https://docs.google.com/document/d/1sitzsboY3oUzjG5SoweLDvTiiQ4A-yLUEOTqdu1U8q8/edit)

### Questions

- We need to agree on #698, #699, #700, #701 before this one can be picked up.

### Tasks to complete

- [ ] Store state tree on DB (materialized path)

Add the following items to the executeTransaction function in transaction.agent.ts

- [x] Translate tx.payload and current bpi account state into appropriate circuit inputs
- [x] Hash(root of the merkelized invoice + Hash (witness))
- [x] Store the hash as the leaf on state tree

ognjenkurtic:
@biscuitdey Do you have any questions or comments related to this one, based on our discussion from the last sync meeting? I am now starting to go through the proposed approach from the linked docs to try to finalize the issue description and acceptance criteria.

ognjenkurtic:
SRI use-case doc proposal:

> DB storage of the BPI Account looks like this:
> 
> 1. stateTree

MerkeTree where a leaf is the root hash of the combination of the root of the merkelized invoice and the hash of the witness, both being part of the output of the executed workstep. 

> 2. historyTree

MerkeTree where each leaf is a root of the state tree, with history tree root being recalculated each time.

> 3. Merkelized payload (Invoice)

Latest state of the payload which is being synchronized (an invoice), merkelized

> 4. Witness

Latest witness as produced by the executed circuit


Open question:

Do we need to keep track (store) points 3 and 4 for every workstep execution or just latest one?

ognjenkurtic:
[Here](https://makandracards.com/makandra/45275-storing-trees-in-databases) is a good overview of patterns that can be used to store a merkle tree in a relational database. Option 2: Materialized Path sounds like something we could utilize unless we decide to go with a JSON blob approach.

Ybittan:
[...]

PR #732, opened 2023-08-28
Feat/bpi account storage prisma poc

ognjenkurtic:
# Description

From POC it turned into a full implementation of state tree leaf storage after the execution of a  transaction. It relies on the previous implementation of merkle tree agent and storage agent to perform marshaling of the state tree to a text field in the DB - merkelized path pattern is not suitable for our current use case as it too complex and we can rely on state tree leaf index to store accompanying data for the leaf. Storage of this data is also implemented as part of the PR. Added a bunch of TODOs for upcoming milestone. Prisma is upgraded to version 5 as part of this PR as well.

## Related Issue

#702 

## Motivation and Context

Because having a BPI without state is like having a kafana without cigarette smoke.

## How Has This Been Tested

Not tested yes. Will be tested as part of #729 with e2e tests and with unit tests as part of milestone 5 once prisma mocking is implemented.

## Types of changes

## Checklist


Therecanbeonlyone1969:
@ognjenkurtic good to review?

ognjenkurtic:
> @ognjenkurtic good to review?

Yes now, i rebased main. @Therecanbeonlyone1969 @Kasshern @biscuitdey

examples/bri-3/src/bri/identity/bpiAccounts/agents/bpiAccountsStorage.agent.ts
```diff
@@ -90,4 +94,20 @@ export class BpiAccountStorageAgent extends PrismaService {
...
     });
   }
+
+  async storeAccompanyingStateLeafValues(
+    id: string,
+    leafIndex: number,
+    merkPayload: MerkleTree,
+    witness: Witness,
+  ): Promise<void> {
+    await this.bpiAccountStateTreeLeafValue.create({
```
biscuitdey, line 104: bpiAccountStateTreeLeafValue class been implemented?

examples/bri-3/prisma/prisma.service.ts
```diff
@@ -1,15 +1,9 @@
-import { Injectable, OnModuleInit, INestApplication } from '@nestjs/common';
+import { Injectable, OnModuleInit } from '@nestjs/common';
 import { PrismaClient } from '@prisma/client';
 
 @Injectable()
 export class PrismaService extends PrismaClient implements OnModuleInit {
   async onModuleInit() {
     await this.$connect();
   }
-
```
Kasshern, line 9: Thank you for removing this. We should also create a nestjs module for prisma. I beleive the way its being injected now we have a different instantiation of prisma for each new connection. If we create the nestjs prisma module, that should no longer be the case.

examples/bri-3/src/bri/identity/bpiAccounts/models/bpiAccount.ts
```diff
@@ -18,21 +19,22 @@ export class BpiAccount {
   stateObjectProverSystem: string;
 
   @AutoMap()
-  stateObjectStorage: string;
+  stateTreeId: string;
```
Kasshern, line 22: A BPI account has a direct line to a single state tree? also why needs id and tree itself?

examples/bri-3/src/bri/merkleTree/agents/import MerkleTree from 'merkletreejs';.ts
```diff
@@ -0,0 +1,20 @@
+import MerkleTree from 'merkletreejs';
```
Kasshern, line 1: @ognjenkurtic Does this file serve a purpose or was it for dev testing?

examples/bri-3/src/bri/merkleTree/agents/merkleTree.agent.ts
```diff
@@ -12,14 +12,11 @@ export class MerkleTreeAgent {
...
-  public createNewMerkleTree(
-    leaves: string[],
-    hashAlgName: string,
-  ): BpiMerkleTree {
+  public createNewMerkleTree(leaves: string[]): BpiMerkleTree {
     return new BpiMerkleTree(
       v4(),
-      hashAlgName,
-      this.service.formMerkleTree(leaves, hashAlgName),
+      'sha256',
```
Kasshern, line 18: @ognjenkurtic if we are sticking to sha256 for everything, might as well create some global hash constant somewhere that makes sense.

examples/bri-3/src/bri/merkleTree/agents/merkleTreeStorage.agent.ts
```diff
@@ -55,6 +55,9 @@ export class MerkleTreeStorageAgent extends PrismaService {
       where: { id: merkleTree.id },
     });
 
+    // TODO: Mil5 - Mapping from a prisma client type to a domain object is not working.
```
Kasshern, line 58: What was not working about this? I added in a mapping within the profile that unmarshalls the tree when its mapped. maybe something similar?

Kasshern (commented):
@ognjenkurtic Left some comments, mostly question maybe one or two suggestions. Should be good to approve after answered.

examples/bri-3/src/bri/identity/bpiAccounts/agents/bpiAccountsStorage.agent.ts
ognjenkurtic: Yes, if you are refering to the prisma model.

examples/bri-3/prisma/prisma.service.ts
ognjenkurtic: I think that by default nestjs provider is a singleton. But i agree we shoudl extract to a separate module, it is cleaner. Could you open a issue for mil5?

examples/bri-3/src/bri/identity/bpiAccounts/models/bpiAccount.ts
ognjenkurtic: Currently yes, my understanding was that it is a bpi account that should store the state for a single workflow. Is that not the case? Regarding the id, i believe it is required because of the automapping setup which is not correct. We can fix this as part of the automapper fixes we ll do in mil5.

examples/bri-3/src/bri/identity/bpiAccounts/accounts.module.ts
```diff
@@ -32,4 +32,5 @@ export const QueryHandlers = [
   ],
   exports: [BpiAccountAgent, BpiAccountStorageAgent],
 })
+// TODO: Move this under a separate folder 'state' instead of 'identity'
```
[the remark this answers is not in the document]
ognjenkurtic: I left it as a reminder as bpiaccounts deal with state storage. Not everything from identity should go into state module, i was just lazy imprecise with my todo. Will open an issue describing the proposal for change.

examples/bri-3/src/bri/merkleTree/agents/merkleTree.agent.ts
ognjenkurtic: Yes, i thing i left a todo somewhere to move this into env. Will be tackled as part of mil5.

examples/bri-3/src/bri/merkleTree/agents/merkleTreeStorage.agent.ts
ognjenkurtic: Your approach works. We need to investigate if it is an optimal approach, since we would be introducing additional classes that represent prisma models just for the sake of mapping. Best options is if we could use prisma models directly in the mapping profiles, but i couldnt get that to work. Mil5 stuff, will be opening an issue

examples/bri-3/src/bri/merkleTree/agents/import MerkleTree from 'merkletreejs';.ts
ognjenkurtic: Ah sorry, will remove it

ognjenkurtic:
@Kasshern @biscuitdey addressed all the  comments i think. Please rereview.

examples/bri-3/src/bri/merkleTree/agents/import MerkleTree from 'merkletreejs';.ts
ognjenkurtic: Removed

biscuitdey (approved):
LGTM

Kasshern (approved):

Issue #743, opened 2023-09-10
SRI - Extract 'sha256' as hash algh used for the merke trees to env


ognjenkurtic:
## Overview

It is currently hardcoded across the codebase. Extract to env to have it in a single place.

### Acceptance

- Value added to env file and env.sample and documented
- All hardcoded occurences replaced with a read from env

## Tasks

- [ ] Do it

djanluka:
assign me

PR #756, opened 2023-09-28
Fixed accounts.controller test for Merkle tree hash algh

djanluka:
# Description

Fixed test for account controller with extracted var for merkle tree hash algh  

## Related issues
[SRI - Extract 'sha256' as hash algh used for the merke trees to env #743
](https://github.com/eea-oasis/baseline/issues/743)

### DIFF
--- a/examples/bri-3/src/bri/identity/bpiAccounts/api/accounts.controller.spec.ts
+++ b/examples/bri-3/src/bri/identity/bpiAccounts/api/accounts.controller.spec.ts
@@ -38,6 +38,10 @@ describe('AccountController', () => {
   let subjectAccountStorageAgentMock: DeepMockProxy<BpiSubjectAccountStorageAgent>;
   let merkleTreeStorageAgentMock: DeepMockProxy<MerkleTreeStorageAgent>;
 
+  beforeAll(async () => {
+    process.env.MERKLE_TREE_HASH_ALGH = 'sha256';
+  });
+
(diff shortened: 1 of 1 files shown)


examples/bri-3/src/bri/identity/bpiAccounts/api/accounts.controller.spec.ts
```diff
@@ -38,6 +38,10 @@ describe('AccountController', () => {
   let subjectAccountStorageAgentMock: DeepMockProxy<BpiSubjectAccountStorageAgent>;
   let merkleTreeStorageAgentMock: DeepMockProxy<MerkleTreeStorageAgent>;
```
Kasshern, line 39: Seems this can be removed.

examples/bri-3/src/bri/identity/bpiAccounts/api/accounts.controller.spec.ts
```diff
@@ -38,6 +38,10 @@ describe('AccountController', () => {
...
 
+  beforeAll(async () => {
```
Kasshern, line 41: works for me as a band-aid. Testing should likely have its own config file/files that set these variables among others.  I believe this work has already begun in builders.ts and testData.helper.ts

Kasshern (approved):

ognjenkurtic (approved):

PR #760, opened 2023-10-23
Build fixes

ognjenkurtic:
# Description

Build was broken because of incorrect usage of hardhat ethers in application source. Hardhat ethers is intended to be used only inside hardhat scripts, tests and tasks. For app purposes, we should use ethers.js

## Related Issue

n\a

## Motivation and Context

Fixes the build

## How Has This Been Tested

local build, start, test and test:e2e

## Types of changes

## Checklist


biscuitdey:
@ognjenkurtic Can you please elaborate on why we can't use hardhat-ethers?
Because the bug I faced was because the `hardhat.config` file was using `es module` while hardhat only supports `commonjs` modules if the project is in typescript. So simply changing the hardhat.config file to commonjs was enough to resolve the problem.

ognjenkurtic:
> @ognjenkurtic Can you please elaborate on why we can't use hardhat-ethers? Adding a `hardhat.config` to our nestjs project makes it also a hardhat project. So we can use the hardhat-etherjs anywhere in the project.
> 
> The bug I faced was because the `hardhat.config` file was using `es module` while hardhat only supports `commonjs` _config file_ if the project is in typescript. So simply changing the hardhat.config file to commonjs was enough to resolve the problem.

This plugin adds an ethers object to the Hardhat Runtime Environment (HRE). You can read more about HRE [here](https://hardhat.org/hardhat-runner/docs/advanced/hardhat-runtime-environment). It is basically a development environment and the hardhat-ethers plugin just makes it easier to use ethers.js in hardhat tasks, scripts and tests.

When your application needs to talk to the blokchain in production, you should directly rely on ethers.js or similar libs as hardhat is not supposed to be present at runtime.

Kasshern (approved):
Getting a Prisma warning to upgrade to 5.5.2. Otherwise Looks good and starts correctly for me.

biscuitdey:
> > @ognjenkurtic Can you please elaborate on why we can't use hardhat-ethers? Adding a `hardhat.config` to our nestjs project makes it also a hardhat project. So we can use the hardhat-etherjs anywhere in the project.
> > The bug I faced was because the `hardhat.config` file was using `es module` while hardhat only supports `commonjs` _config file_ if the project is in typescript. So simply changing the hardhat.config file to commonjs was enough to resolve the problem.
> 
> This plugin adds an ethers object to the Hardhat Runtime Environment (HRE). You can read more about HRE [here](https://hardhat.org/hardhat-runner/docs/advanced/hardhat-runtime-environment). It is basically a development environment and the hardhat-ethers plugin just makes it easier to use ethers.js in hardhat tasks, scripts and tests.
> 
> When your application needs to talk to the blokchain in production, you should directly rely on ethers.js or similar libs as hardhat is not supposed to be present in production.

Understood.

biscuitdey (approved):
Will add an issue for implementing etherjs functions.

Issue #740, opened 2023-09-10
SRI - Fix mapping from Prisma models to domain objects


ognjenkurtic:
## Overview

Currently automapping from Prisma models to domain objects is broken, as automapper complains about using type (Prisma model) when invoking the mapping so we use domain objects as source which causes various issues and worksarounds (look for TODOs with the id of this issue).

example of a wrong mapping:

`return this.mapper.map(bpiAccountModel, BpiAccount, BpiAccount);`

should be something like:

`return this.mapper.map(bpiAccountModel, BpiAccountPrismaClientModel, BpiAccount);`

### Reference

https://github.com/nartc/mapper/discussions/280

### Questions

- Do we to introduce a new type of class (Prisma Model DTO) that only holds the properties of the PrismaModel thus making automapper happy (look at MerkleTreeDto in merkle.profile.ts)? What are the drawbacks if we use this approach?
- Can we stick with PrismaModels but somehow configure Automapper to like them? i.e. [pojo strategy](https://www.npmjs.com/package/@automapper/pojos)

### Acceptance

- mapping is performed by giving automapper either a prisma model (type) or a new object which represenets the model (class).
- All mappings in the app are fixed according to the new approach
- New approach is documented in the readme 
- All tests pass
- 
## Tasks

- [x] Research and decide on the approach
- [x] Implement POC
- [x] Apply everyhwere
- [x] Update readme

ognjenkurtic:
This branch has a POC of using the pojos strategy of automapper to achieve mapping from prisma model to a domain object, on the example of reading a bpi subject via an id

feature/740-fix-mappings-from-prisma-models-to-domain-objects

Issue #769, opened 2023-11-23
Fix errors in e2e tests


skosito:
Running e2e tests has 2 problems:

1. error message even though tests pass
`Failed execution of transaction with id <uuid>. Error: NotImplementedException: Not Implemented`

This message should not appear.

2. hangs after execution, currently fixed with `--forceExit`

Ideally we should not use this flag.

ognjenkurtic:
@biscuitdey could you please open the remaining issues that are necessary to make the e2e test green?

PR #775, opened 2024-01-04
Feature/740 fix mappings from prisma models to domain objects

ognjenkurtic:
# Description

Implement an example of prisma model -> domain object mapping that does not rely on automapper as the automapper pojos strategy is bolierplate heavy and does not solve some edge cases. 

New approach uses Object.assign and requires a single method per model - domain object pair. 

If this approach looks good, i ll apply it across the  codebase in a separate PR.

## Related Issue

#740 

## Motivation and Context

Mapping was not done properly 

## How Has This Been Tested

Tested locally with postman, trying to fetch BpiSubject by id with associated roles and verified all mappings correctly performed.
Ran all unit tests

## Types of changes

## Checklist


examples/bri-3/prisma/prisma.mapper.ts
```diff
@@ -0,0 +1,38 @@
+import { Injectable } from '@nestjs/common';
+import { BpiSubject as BpiSubjectPrismaModel, BpiSubjectRole as BpiSubjectRolePrismaModel } from '@prisma/client';
+import { BpiSubject } from '../src/bri/identity/bpiSubjects/models/bpiSubject';
+import { BpiSubjectRole } from '../src/bri/identity/bpiSubjects/models/bpiSubjectRole';
+import { IDomainObject } from '../src/shared/domainObject.interface';
+
+// We do mapping from prisma models to domain objects using simple Object.assing
```
Kasshern, line 7: assing should be assign

examples/bri-3/src/bri/identity/bpiSubjects/models/bpiSubject.ts
```diff
@@ -1,8 +1,9 @@
 import { AutoMap } from '@automapper/classes';
 import { v4 } from 'uuid';
+import { IDomainObject } from '../../../../shared/domainObject.interface.js';
 import { BpiSubjectRole } from './bpiSubjectRole';
 
-export class BpiSubject {
+export class BpiSubject implements IDomainObject  {
```
Kasshern, line 6: why is it necessary to have the implemented interface IDomainObject and private activator<T extends IDomainObject>. Just to ensure a common inherited value of id?

Kasshern (approved):
I agree with this approach. It reflects the need for some amount of custom work to be done, which I think we should not shy away from at this point in the development process.  Left a question.

Kasshern:
@ognjenkurtic please clear conflicts.

examples/bri-3/src/bri/identity/bpiSubjects/models/bpiSubject.ts
ognjenkurtic: The interface is not needed at all. I was following an article and thought that it might be a good idea in order to make sure only domain objects are accepted for activation but on second thought i think it is over engineering. Removed.
The activator is necessary to avoid instantiation of each domain object ctor by providing empty params. I was hoping for a fully generic solution (one method for any mapping pair) but couldn't make it work so stopped here for now.

ognjenkurtic:
> @ognjenkurtic please clear conflicts.

@Kasshern done and addressed all other comments, please rereview

Kasshern:
@ognjenkurtic e2e test failing

ognjenkurtic:
e2e step is failing because of this https://github.com/eea-oasis/baseline/issues/769

Kasshern (approved):

Therecanbeonlyone1969 (commented):
LGTM

biscuitdey (approved):
Approach looks good to me.

Issue #776, opened 2024-01-05
SRI - Implement schema parser for translating tx payload into circom inputs 


ognjenkurtic:
## Overview

Implement the parser based on the description in the [design document](https://docs.google.com/document/d/1AzyUK0OMATe0XRPaTbayRR3KmuaFLMAKybIytaVVBdY/edit). It should support all the cases  present there. 

Should be implemented as a standalone stateless service that accepts a schema object in JSON and produces an object representing  circom inputs ready for consumption.

Make sure to write unit tests for various schema use-cases (nested fields, various data types, default values, errors etc.)

### Reference

https://github.com/eea-oasis/baseline/issues/744
https://github.com/eea-oasis/baseline/issues/766

### Acceptance

- Parser implemented as a standalone service
- Supports all the cases present in the deisgn doc example
- 100% test coverage

## Tasks

- [x] Do it

Ybittan:
ZK Circuits Input Transformer

PR #778, opened 2024-01-08
Feat/776 circuit inputs parser

ognjenkurtic:
# Description

Introduces a first version of a circuit input parser that translates tx payload into circuit input using a predefined schema.
Adds a number of unit tests to cover the basic functionality

## Related Issue

#776 

## Motivation and Context

Allows the BPI to support general use-cases when it comes to circuits. Workgroup admin would provide the compiled circuits together with the schema, which would allow the BPI to parse the tx payload and execute the circuits

## How Has This Been Tested

Introduced circuitInputParser.service.spec.ts with 10 new unit tests, covering most of the new functionality

## Types of changes

## Checklist

### DIFF
--- a/examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/circuitInputsParser/circuitInputParser.service.spec.ts
+++ b/examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/circuitInputsParser/circuitInputParser.service.spec.ts
@@ -25,6 +25,7 @@ describe('CircuitInputsParserService', () => {
       `{
         "supplierInvoiceID": "INV123"
        }`;
+
     const schema = {
       mapping: [{
         circuitInput: 'dascircuitinput',
@@ -47,13 +48,16 @@ describe('CircuitInputsParserService', () => {
       `{
         "somethingElse": ""
        }`;
+
+    const defaultParamValue = '555333';
+
     const schema = {
       mapping: [{
         circuitInput: 'dascircuitinput',
         description: 'desc',
         payloadJsonPath: 'supplierInvoiceID',
         dataType: 'string',
-        defaultValue: '555333'
+        defaultValue: defaultParamValue
       } as CircuitInputMapping]
     } as CircuitInputsMapping;
 
@@ -61,7 +65,7 @@ describe('CircuitInputsParserService', () => {
     const circuitInputs = cips.applyMappingToJSONPayload(payload, schema);
 
     // Assert
-    expect(circuitInputs).toStrictEqual({ "dascircuitinput": "555333" });
+    expect(circuitInputs).toStrictEqual({ "dascircuitinput": defaultParamValue });
   });
 
   it('Should return null based on a missing default value for a missing string param at root level', () => {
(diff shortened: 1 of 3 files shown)


examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/circuitInputsParser/circuitInputParser.service.ts
```diff
@@ -0,0 +1,95 @@
...
+      if (currentValue[part] === undefined) {
+        return undefined;
+      }
+      currentValue = currentValue[part];
+    }
+
+    return currentValue;
+  }
+
+  private calculateStringCharCodeSum(text: string): number {
```
skosito, line 86: maybe this method can be more readable with usage of `.map` and things like that? `text` can be transformed into array using `.split` method and then `.map` can be used or something like that

but not super important

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/circuitInputsParser/circuitInputParser.service.ts
```diff
@@ -0,0 +1,95 @@
...
+    try {
+      const jsonPayload = JSON.parse(payload);
+
+      for (let mapping of cim.mapping) {
+        const value = this.getJsonValueByPath(
+          jsonPayload,
+          mapping.payloadJsonPath,
+        );
+
+        if (!value && !mapping.defaultValue) {
```
skosito, line 20: should we check value === undefined, if value can be empty string? since we are checking ===undefined in method itself

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/circuitInputsParser/circuitInputParser.service.ts
```diff
@@ -0,0 +1,95 @@
...
+          this.logger.logError(
+            `Missing value and default value for mapping ${cim.mapping} while mapping circuit inputs for payload ${payload}`,
+          );
+          return null;
+        }
+
+        switch (mapping.dataType) {
+          case 'string':
+            result[mapping.circuitInput] = value
```
skosito, line 29: can this be simplified, like result... = this.calc(value || default)

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/circuitInputsParser/circuitInputParser.service.spec.ts
```diff
@@ -0,0 +1,259 @@
...
+       }`;
+
+    const schema = {
+      mapping: [
+        {
+          circuitInput: 'dascircuitinput',
+          description: 'desc',
+          payloadJsonPath: 'supplierInvoiceIDs',
+          dataType: 'array',
+          arrayType: 'string',
```
skosito, line 248: do we need both array and arrayType? can arrayType be calculated using `typeof array[0]`? can `typeof value` be used in other places too to simplify mapping logic and remove some fields here?

i feel more validation of this object is needed, for example if dataType is not array, we should not specifiy arrayType and so on, but it seems like some simplification can be done first

ognjenkurtic: We want the participants to strictly define the types that are expected in the payload, so that we can do a type check before we even try to invoke the circuit. there is still definitely room for improvement, both for simplification and additional validation, but let's do it in small iterations as we move forward.

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/circuitInputsParser/circuitInputParser.service.ts
ognjenkurtic: i like the good old for loop here and there

skosito (approved):

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/circuitInputsParser/circuitInputParser.service.spec.ts
```diff
@@ -0,0 +1,259 @@
+import { DeepMockProxy, mockDeep } from 'jest-mock-extended';
```
Therecanbeonlyone1969, line 1: @ognjenkurtic this is a test not a service ... ????

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/circuitInputsParser/circuitInputParser.service.spec.ts
Therecanbeonlyone1969: agree. the `circuitInput` for example will be an object that has a schema itself depending on the type of input parameters.

examples/bri-3/src/bri/zeroKnowledgeProof/services/circuit/circuitInputsParser/circuitInputParser.service.ts
```diff
@@ -0,0 +1,93 @@
+import { Injectable } from '@nestjs/common';
+import { LoggingService } from '../../../../../shared/logging/logging.service';
+
+@Injectable()
+export class CircuitInputsParserService {
+  constructor(private readonly logger: LoggingService) {}
+
+  public applyMappingToJSONPayload(payload: string, cim: CircuitInputsMapping) {
```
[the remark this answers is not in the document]
Therecanbeonlyone1969: @ognjenkurtic I do not like what you did. How about we use this? ... https://github.com/grantila/suretype  ... looks easy to implement and very flexible when the user provides the schema.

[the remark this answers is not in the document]
ognjenkurtic: Agreement is not to use the library per the last SRI sync call. Will address other comments if any and submit for rereview

ognjenkurtic:
@Therecanbeonlyone1969 all comments are addressed now. Please approve so that i can merge.

Therecanbeonlyone1969 (approved):
Lgtm now

PR #780, opened 2024-01-11
fix mappings from prisma models to domain objects accross the codebase

ognjenkurtic:
# Description

Apply the previously approved POC for mapping of prisma models to domain objects across the codebase.
PR review hint: Go through commits one by one from oldest.

## Related Issue

#740 

## Motivation and Context

Fixes broken mapping that was present previously 

## How Has This Been Tested

Ran all unit tests
Ran e2e test locally
Executed random endpoints  via Postman manually

## Types of changes

## Checklist

- [] I have added tests to cover my changes.


examples/bri-3/src/shared/prisma/prisma.mapper.ts
```diff
@@ -0,0 +1,161 @@
...
+
+interface IConstructor<T> {
+  new (...args: any[]): T;
+}
+
+@Injectable()
+export class PrismaMapper {
+  constructor(private readonly merkleTreeService: MerkleTreeService) {}
+
+  public mapBpiSubjectPrismaModelToDomainObject(
```
skosito, line 43: im a bit rusty with typescript, but can these be like generic method, and then only define these verbose methods if there is something more tricky than simple assign? generic like this (this is also ugly, but just as illustration):

```
public map<T extends object>(source: any, targetType: IConstructor<T>): T {
    const target = this.activator(targetType)
    Object.assign(target, source);
    return target;
  }

```

ognjenkurtic: Thanks for this, spent some time during POC trying to implement the generic method but got stuck with some dumb issue and decided to leave it for later. Applied now everywhere based on your proposal, please rereview. We can improve as we go. @skosito

skosito (approved):

Therecanbeonlyone1969 (approved):
LGTM

Issue #742, opened 2023-09-10
Implement mocking of the Prisma client for unit testing purposes


ognjenkurtic:
## Overview

Currently there is too  much bolierplate when bootstraping a unit test, as we are mocking every single storage agent. We can avoid this by mocking the prisma client and having it setup as a kind of in-memory database. Scope fo this one is to research https://github.com/demonsters/prisma-mock or any other prisma mocking library and, if suitable, implement a POC for a single set of unit tests (you can find candidates by searching for this issue id in the TODO comments) 

### Reference

- https://github.com/demonsters/prisma-mock

### Questions

- Does this libary cover all the Prisma features we are using?
- Are there other libs we could use? 

### Acceptance

- Lib validated
- POC on a single set of unit tests implemented

## Tasks

- [ ] Do it

ognjenkurtic:
@skosito could you please open issue to apply the changes from #786 in other places?

PR #786, opened 2024-01-29
Introduce prismock and use it in transaction agent spec

skosito:
# Description

- Adds [prismock library](https://www.npmjs.com/package/prismock). I used this instead of [prisma-mock library](https://www.npmjs.com/package/prisma-mock) which is mentioned in issue because it seems maintained more often, and they are doing the same thing anyways.

- Reworks transaction agent specs to use this library and fix some tests there. NOTE: make sure to expand diff in transaction agent spec since it's hidden in diff by default.

- Bumps nodejs version to `17.0.0` in workflow because [prisma needs it](https://bobbyhadz.com/blog/structured-clone-is-not-defined-in-javascript#solved-structuredclone-is-not-defined-in-js-and-nodejs).

- Cleanup of `NOT_FOUND` messages in workstep and workflow modules.

- Once approach is approved I will modify all unit tests in the repo in follow up PR.
## Related Issue

#742 

## Motivation and Context

Better way to write unit tests.

## How Has This Been Tested

This modifies tests only, so it is tested by default. 

## Types of changes

## Checklist


examples/bri-3/src/bri/transactions/agents/transactions.agent.spec.ts
```diff
@@ -1,330 +1,339 @@
-import { DeepMockProxy, mockDeep } from 'jest-mock-extended';
-import { uuid } from 'uuidv4';
-import { ICircuitService } from '../../../bri/zeroKnowledgeProof/services/circuit/circuitService.interface';
-import { TestDataHelper } from '../../../shared/testing/testData.helper';
 import { AuthAgent } from '../../auth/agent/auth.agent';
-import { BpiSubjectAccount } from '../../identity/bpiSubjectAccounts/models/bpiSubjectAccount';
-import { BpiSubject } from '../../identity/bpiSubjects/models/bpiSubject';
+import { BpiSubjectAccount as ModelBpiSubjectAccount } from '../../identity/bpiSubjectAccounts/models/bpiSubjectAccount';
```
ognjenkurtic, line 2: it is the other way around in the prisma mapper when it comes to naming.
Domain objects imported without aliases.
PrismaModels have PrismaModel suffix.
Could we do the same here and use this as a convention throughout the codebase?

We can also turn it around, i do not care as long as we are consistent

examples/bri-3/src/bri/transactions/agents/transactions.agent.spec.ts
```diff
@@ -1,330 +1,339 @@
...
-const merkleTreeServiceMock: DeepMockProxy<MerkleTreeService> =
-  mockDeep<MerkleTreeService>();
-const circuitsServiceMock: DeepMockProxy<ICircuitService> =
-  mockDeep<ICircuitService>();
-
-// TODO: #742 Setup of this test data below is what should be handled in a separate file where we mock only prisma.client
-// and implement various test data scenarios that can be selected with a single line of code.
-// https://github.com/demonsters/prisma-mock
-const existingWorkgroupId = uuid();
+  bpiSubject2 = await prisma.bpiSubject.create({
```
ognjenkurtic, line 84: we have a lot of builiders implemented in builders.ts.
Could we can use them to define the test domain objects we want to work with and then just have separate methods that abstract away prisma.create calls? This way we can share those methods across the test code base.

This way we could probably ditch the as ModelBpiSubjectAccount syntax as well.

Convetion would be that we work with domain objects in tests, and prisma create calls are only there to avoid mocking the storage agents

examples/bri-3/src/bri/transactions/agents/transactions.agent.spec.ts
```diff
@@ -1,330 +1,339 @@
...
-    );
-
     const tx = new Transaction(
       '1',
       1,
-      '123',
-      '123',
-      fromBpiSubjectAccount,
-      undefined as unknown as BpiSubjectAccount,
+      workflow.id + '',
```
ognjenkurtic, line 230: why the + ''?

examples/bri-3/src/bri/transactions/agents/transactions.agent.spec.ts
```diff
@@ -1,330 +1,339 @@
...
     // Arrange
-    workflowStorageAgentMock.getWorkflowById.mockResolvedValueOnce(
-      existingWorkflow1,
-    );
-
-    workstepStorageAgentMock.getWorkstepById.mockResolvedValueOnce(
-      existingWorkstep1,
-    );
-
     const tx = new Transaction(
```
ognjenkurtic, line 227: could you add transaction builder as well?

examples/bri-3/src/bri/transactions/agents/transactions.agent.spec.ts
ognjenkurtic: It is mainly for readability and maintainability purposes. I agree it might be too much at this point. Maybe just open an issue and we can introduce them later if time allows.

Therecanbeonlyone1969: @skosito can you open an issue for this so we can keep track of it? cc @ognjenkurtic @Kasshern

Therecanbeonlyone1969 (approved):
Modulo the builder comment from @ognjenkurtic which will be addressed in a separate PT I am good

biscuitdey (approved):
Approach looks good to me

examples/bri-3/src/bri/transactions/agents/transactions.agent.spec.ts
ognjenkurtic: @skosito dont forget to open an issue for this one pls.

Issue #815, opened 2024-07-18
SRI - Add the Verifirer.sol as an zkp artifact to the BPI and env


ognjenkurtic:
- [x] Do it
- [ ] Extend readme
- [x] Deploy verifier contract in deploy.ts

ognjenkurtic:
@biscuitdey what is the status here? are we going with 3 separate verifier contracts or a single one?

biscuitdey:
> @biscuitdey what is the status here? are we going with 3 separate verifier contracts or a single one?

@ognjenkurtic We are using 3 separate verifier contracts for the 3 worksteps for now, 
because the verification key values are hardcoded inside the contract.
Using one verifier contract (where we pass the verification key as a function
param) is not possible as the contract uses assembly inside solidity contract
and assembly does not accept local variables. Also, if we declare the values as
storage variables, then we would have to change the verifyFunction code it as
there is different syntax to use storage variables.

Andreas suggested that we create a single verifier contract as an improvement later on. This would require us to alter how the verification key variables are called and utilised. Do you suggest we do it now or keep it for later?

ognjenkurtic:
I think contract per workstep is fine for now, we can stick with 3

PR #818, opened 2024-07-18
SRI - 815 Add verifier.sol for workstep circuits

biscuitdey:
# Description

This PR adds the verifier contracts for the workstep circuits.
The verifier contracts are generated while compiling the circuits using the script (plonk.sh).

We are using three separate verifier contracts for the 3 worksteps 
because the verification key values are hardcoded inside the contract. 
Using one verifier contract (where we pass the verification key as a function
param) is not possible as the contract uses assembly inside solidity contract 
and assembly does not accept local variables. Also, if we declare the values as
storage variables, then we would have to change the verifyFunction code it as 
there is different syntax to use storage variables.

## Related Issue

#815 

## How Has This Been Tested

Tests would be added in upcoming PRs.

## Types of changes

## Checklist


biscuitdey:
@Therecanbeonlyone1969 Do we need all three workstep circuit contracts? OR can we use only the workstep1 verifier contract to verify all three proofs (workstep1, workstep2 and workstep3)?

ognjenkurtic:
> @Therecanbeonlyone1969 Do we need all three workstep circuit contracts? OR can we use only the workstep1 verifier contract to verify all three proofs (workstep1, workstep2 and workstep3)?

My understanding from the call is that we can use a single one, although not sure why would the command create one contract for each circuit. I see that they differ only in the first dozen const variables. Anyways, leaving to @Therecanbeonlyone1969 for review.

Therecanbeonlyone1969 (dismissed):
LGTM ... are we doing 3 contracts because of the different verification keys by circuit? @biscuitdey

biscuitdey:
> LGTM ... are we doing 3 contracts because of the different verification keys by circuit? @biscuitdey

@Therecanbeonlyone1969 Yes. The verification key values (Qm, Ql, etc.) are hardcoded in each of the verifier contract.

biscuitdey:
@ognjenkurtic @Therecanbeonlyone1969  I created a single verifier contract that accepts verification key as a param. Please re-approve PR.

Therecanbeonlyone1969 (approved):
LGTM. As an improvement, next time, let's update the verifier code to take storage variables

